PR #128, opened 2017-03-17
Dev to master comparison

ManjotS:
may require squash merge as discussed @navdeepsinghkhalsa


assets/css/style.css
```diff
@@ -301,12 +332,21 @@ footer {
 /**************
  * Shabad Page
  **************/
+blockquote {
```
navdeepsinghkhalsa: Are blockquotes used outside of .search-results? Why are there two separate rules?

ManjotS: search results and shabad.. also since its an override of foundation, I didn't want to effect expected use in any other cases

assets/css/style.css
```diff
@@ -357,6 +397,7 @@ footer {
 }
 #metadata {
   border: 1px solid #cacaca;
+
```
navdeepsinghkhalsa: Need to watch out for unneeded whitespace. Will be implementing coding style guide soon.

ManjotS:
doesn't min take care of that? i prefer readablility

inc/footer.php
```diff
@@ -12,31 +12,44 @@
...
 ?>
+<!-- preloader -->
+<div class="hidden">
+  <script type="text/javascript">
+    <!--//--><![CDATA[//><!--
 
+      if (document.images) {
+        preborder1 = new Image();
+
+        preborder1.src = "/assets/images/border1.png";
```
navdeepsinghkhalsa: Why are we preloading this?

ManjotS: yeah we dont need it anymore

ManjotS: i just wanted to preserve the preload code for future use is the main reason for not pulling it out

navdeepsinghkhalsa: Preloading will not be necessary with the direction in which this site is going and code can be looked up again if needed. Sorry, but could you please remove and update the PR?

PR #186, opened 2017-11-07
Added arrows to metadata at top and added pagination to shabad container

ManjotS:
takes advantage of api navigation data fix #94

PR #187, opened 2017-11-07
Use JSX over h

bogas04:
## What ?

Since we use `h()` for most of our DOM creation, we can easily port our code-base to use [jsx](https://facebook.github.io/jsx/). 

## Why ?

The value out of that is that we can then also move to supported frameworks, namely React-like ([React](https://reactjs.org/)/[Preact](https://preactjs.com/)/[Inferno](https://infernojs.org/)) and even [VueJS](https://vuejs.org/v2/guide/render-function.html#JSX). This will help us clean up our code-base, and might also encourage contributions from Open Source volunteers. This isn't only about framework though, since we would be moving to [webpack](https://webpack.js.org/) as our build process, and discouraging global mutations in favour of modularity, it just means that our code would be extensible and easily maintainable. We can also pre-compile gzip and brotli files at build time, saving giving us a boost compared to on-the-fly compression as well as wins from brotli (20%).

## How ?

Phase 1 (this PR):
* [x] *mostly done* migrate from `h()` to `JSX` with use of `babel-plugin-transform-react-jsx`.
* [x] *mostly done* restructure codebase to use `webpack`. Abstract out components and utils.

Phase 2:
* [ ] refactor remaining code that relies on global variables, DOM mutations via jQuery.

Phase 3:
* [ ] migrate to usage of `ReactJS`, or whatever framework we all agree upon. It will become very easy with Phase 1 & 2 completed.


bogas04:
This PR would conflict with #186 due to usage of `baani` in that one.

src/js/components/SearchForm.js
```diff
@@ -0,0 +1,74 @@
+import { entries } from '../util';
+import Option from './Option';
+
+export default ({ }) => {
+  const $search = (
+    <input
+      name="q"
+      id="search"
+      class="gurbani-font"
```
navdeepsinghkhalsa: className

src/js/renderShabad.js
```diff
@@ -1,33 +1,34 @@
...
   document.body.classList.remove("loading");
-  let footnav = '';
+
+  let footnav = null;
   if(typeof nav != "undefined") {
     let link        = navLink(nav);
     let pagination  = [];
     if (typeof nav.previous != "undefined") {
-      pagination.push(h('a', { href: link + nav.previous }, '&laquo;'));
+      pagination.push(<a href={link + nav.previous}>&laquo;</a>);
```
navdeepsinghkhalsa: Does this work? Or do you need to use template literals?

bogas04: Yeah this will work ! :D

src/js/components/SearchForm.js
bogas04: Nice catch, however class will work for our implementation of `h`. It'll also work with `preact` and `react@^16`. So for now we can live with it.

bogas04:
I'll resolve the conflicts again by EoD :)

PR #191, opened 2018-01-08
Split View Feature (fix #87)

### LINKED ISSUE #87: Add split view
The old site had a toggle for split view where you could read (and copy paste) the gurmukhi in a block with each other part in its own block for full shabad/ang/etc

bogas04:
* Add Split View Feature
* Code cleanup
* Transform renderControls to JSX
* Fix a bug where shared panktee was undefined
* Update webpack, node-sass

### DIFF
--- a/src/js/main.js
+++ b/src/js/main.js
@@ -9,6 +9,29 @@ const default_prefs = {
   'shabadToggles': []
 };
 
+function getPrefs() {
+  $.each(default_prefs, function(key, defaults) {
+    prefs[key] = window.localStorage[key] ? JSON.parse(window.localStorage[key]) : defaults;
+  });
+}
+function setPref(key, val) {
+  window.localStorage[key] = JSON.stringify(val);
+}
+function checkboxPref(e, key, option) {
+  if (e.target.classList.contains('active')) {
+    if (prefs[key].indexOf(option) < 0) {
+      prefs[key].push(option);
+    }
+  } else {
+    const index = prefs[key].indexOf(option);
+    prefs[key].splice(index, 1);
+  }
+  setPref(key, prefs[key]);
+}
+
+// IE 10 doesn't support toggle class
+const toggleClass = ($el, cl) => $el.classList[$el.classList.contains(cl) ? 'remove' : 'add'](cl);
+
 getPrefs();
 
 if ($searchType) $searchType.addEventListener("change", updateSearchLang);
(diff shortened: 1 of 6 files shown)


src/js/main.js
```diff
@@ -9,6 +9,29 @@ const default_prefs = {
...
+    }
+  } else {
+    const index = prefs[key].indexOf(option);
+    prefs[key].splice(index, 1);
+  }
+  setPref(key, prefs[key]);
+}
+
+// IE 10 doesn't support toggle class
+const toggleClass = ($el, cl) => $el.classList[$el.classList.contains(cl) ? 'remove' : 'add'](cl);
```
navdeepsinghkhalsa: This is fine for now, but we are only supporting back to IE11. Any system that supports IE10 also supports IE11.

navdeepsinghkhalsa: On further review, it seems that this is used extensively. I would rather we kept `el.classList.toggle`

bogas04:
IE11 doesn't add support for toggle either, so we need to keep this change in order to support IE11.

Once we actually migrate to a UI framework, most of this code would disappear.

screenshot_20180109-092007

navdeepsinghkhalsa:
I saw that on MDN as well, but this page on Microsoft's Developer site indicates that IE11 does support it. It doesn't support the second parameter, but we're not using that anyway.

https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/11865865/

bogas04:
Oh, that's interesting. Alright then, I'll use classList.toggle

src/js/main.js
```diff
@@ -1,77 +1,114 @@
...
   el.show().select();
-  document.execCommand("copy");
+  document.execCommand('copy');
   el.blur().hide();
 });
-$("#shabad").on("click", ".share .twitter", function() {
-  let tweet = $(this).parents(".line").children("textarea").val();
-  let shortURL = '\n'+shortenURL();
+
+$('#shabad').on('click', '.share .twitter', function () {
```
bogas04: Only this part is left in jQuery, I believe.
I'm not sure what the 2nd arg to `on` does, and how it translates to `querySelector`.

```js
  document
    .querySelector('#shabad')
    .addEventListener('click', e => {
      const $node = e.currentTarget;
      if ($node.parentNode.classList.contains('share') && $node.classList.contains('twitter')) {
        // do the thing?
      }
    });
```

navdeepsinghkhalsa: `querySelector` equivalent would be `’#shabad .share .twitter’`

bogas04: I think I'll take up this change in next sprint as the details within this click listener also makes use of some jQuery features that aren't that trivial to port. Can we go ahead with this for now?

navdeepsinghkhalsa: Sounds good. I will review again during my lunch break

PR #192, opened 2018-01-12
Fixed prev/next shabad UI according to Aman Singh's suggestions

ManjotS:
Please review and merge


navdeepsinghkhalsa:
Thank you for your work, @ManjotS. Could you please look into proper nesting for SCSS? It makes for cleaner code.

navdeepsinghkhalsa:
I would suggest to use a border with a 50% radius to make a circle around a single character rather than layering characters

PR #193, opened 2018-01-12
Fix #179 added first 3 letters of hostname to http headers

### LINKED ISSUE #179: Add first 3 digits of hostname as html comment at end
ie if hostname is inw-ded.khalis.net, html should have `<!--inw-->`

ManjotS:
used headers instead of passing vars to userspace

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -27,3 +27,4 @@ app.get('*', function (req, res) {
 // Listen
 const port = process.env.NODE_ENV === 'development' ? '8081' : '8080';
 app.listen(port, () => console.log(`Server started on port:${port}`));
+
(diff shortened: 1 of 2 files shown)


navdeepsinghkhalsa:
Could you please run `npm test` before committing? There are many issues with code style

ManjotS:
Ah good to know. The only file I changed was index.js for this PR. I just submitted the fix.

navdeepsinghkhalsa:
Hmm. I am sorry, but it seems that I neglected to have the test run on the server script; there are still errors. I will update that, but your editor should be catching these. Which editor are you using? Do you have have the linting plugins installed? There's a page on Confluence that lists the plugins for Sublime, Atom, and VSCode. I will add to CONTRIBUTING.md as well.

PR #194, opened 2018-01-12
use more standard css for next/prev

ManjotS:
using fontawesome now


navdeepsinghkhalsa:
Could you please take a look at smaller screen sizes? The alignment is a bit off.
screen shot 2018-01-12 at 6 02 58 pm

ManjotS:
looks ok to me..

PR #213, opened 2018-01-17
Fix event listeners for menu/copy/tweet

bogas04:
* Moved copy/tweet event listeners to renderShabad, where they should
reside.
* Added eventListeners to a DOMReady callback.
* Fixed a typo in code where event listeners weren't attached to element
but window object itself.

### DIFF
--- a/src/js/renderShabad.js
+++ b/src/js/renderShabad.js
@@ -2,9 +2,9 @@ function renderShabad(gurbani, nav) {
   document.body.classList.remove("loading");
 
   let footnav = null;
-  if(typeof nav != "undefined") {
-    let link        = navLink(nav);
-    let pagination  = [];
+  if (typeof nav != "undefined") {
+    let link = navLink(nav);
+    let pagination = [];
     if (typeof nav.previous != "undefined") {
       pagination.push(<div class="shabad-nav left"><a href={link + nav.previous}><i class="fa fa-chevron-left" aria-hidden="true"></i><span>Previous</span></a></div>);
     }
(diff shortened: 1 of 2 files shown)


src/js/renderShabad.js
```diff
@@ -91,6 +91,27 @@ function metaData(data, nav) {
...
 
 function Baani(gurbani) {
+  const onCopyClick = ({ currentTarget }) => {
+    const textarea = currentTarget.parentNode.parentNode.querySelector('textarea');
+    textarea.style.display = 'block';
+    textarea.focus();
+    textarea.select();
+    document.execCommand('copy');
+    textarea.blur();
+    textarea.style.display = 'none';
```
bogas04: Not sure if changing `style.display` is needed.

src/js/renderShabad.js
```diff
@@ -91,6 +91,27 @@ function metaData(data, nav) {
...
+  };
+
+  const onTweetClick = ({ currentTarget}) => {
+    let tweet = currentTarget.parentNode.parentNode.querySelector('textarea').value;
+    const shortURL = `\n${shortenURL()}`;
+    if (tweet.length + shortURL.length > 134) {
+      tweet = `${tweet.substring(0, 132 - shortURL.length)}..`;
```
bogas04: @navdeepsinghkhalsa can we move this to 280 - 8 now, as twitter has increased the limit.

src/js/renderShabad.js
navdeepsinghkhalsa: I believe that was put in because it otherwise didn't perform the select+copy

src/js/renderShabad.js
navdeepsinghkhalsa: Yes, please. Will await that change before merging

bogas04:
Yeah I'll have a look. The way we manage event listeners isn't the nicest way. So it's about time we actually fix it.

PR #228, opened 2018-01-25
Fix keyboard attempt

tarunsingh5:


tarunsingh5:
this should be good to review now - ptal

Issue #181, opened 2017-11-05
Source T is not recognized


ManjotS:
Please ensure Bhai Nand Lal Jee Thankhanama is recognized.

ManjotS:
https://github.com/bogas04/khajana-js/pull/2

ManjotS:
source T is now source R

bogas04:
Is it still an issue?

ManjotS:
works http://dev.sikhitothemax.org/shabad?id=31000&q=pkn&&&source=R&highlight=402704

PR #229, opened 2018-01-25
Fix 99 random url

### LINKED ISSUE #99: /random should just redirect to random shabad
This will ensure:
1) we are not maintaining 2 shabad viewing interfaces (one for hukam, one for shabad/ang)
2) everything we update on shabad (ie showing author etc) is automatically there for random hukam

Don't use 301 redirects .. we just want it to be an http refresh

ManjotS:
fixes #99 rewrites /random to shabad?id=xyz


ManjotS:
@navdeepsinghkhalsa we need the new shabados api (still called khajana-js on npm) to fix #181

Issue #211, opened 2018-01-17
on screen keyboard: biharee closes and breaks keyboard


ManjotS:
## What went wrong ?
clicking this closes and breaks the keyboard requiring reload ![image](https://user-images.githubusercontent.com/644885/35023724-c4f20e2e-faf0-11e7-9ed6-3ac186ba828b.png)

navdeepsinghkhalsa:
Fixed by PR #232

PR #232, opened 2018-01-25
Update renderGurmukhiKeyboard.js

tarunsingh5:
should be ready to merge

### DIFF
--- a/src/js/renderGurmukhiKeyboard.js
+++ b/src/js/renderGurmukhiKeyboard.js
@@ -16,14 +16,10 @@ function renderGurmukhiKeyboard($search) {
         $search.value = $search.value.substring(0, $search.value.length - 1);
         break;
       }
-      case 'page-1': {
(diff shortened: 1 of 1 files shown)


tarunsingh5:
#211 gets fixed by this

bogas04:
Awesome.

Could you please take time to convert that minor jQuery code to vanilla JS one?

```diff
- $("#abc").hide();
+ document.querySelector("#abc").style.display = 'block';
```

We would lose some animation but it'll help us in future while porting.

@navdeepsinghkhalsa what do you think?

tarunsingh5:
@bogas04  - that actually breaks the keyboard - i think we should get it fixed first and then we can optimize later

bogas04:
Sure!

navdeepsinghkhalsa:
I agree with @bogas04. We should avoid technical debt.

tarunsingh5:
Ok, I can't get it to work the way @bogus04 suggested. Please edit.

Issue #216, opened 2018-01-20
Raag Name returns as "False", should return as blank


tarunsingh5:
<img width="671" alt="screen shot 2018-01-19 at 5 00 22 pm" src="https://user-images.githubusercontent.com/35271830/35178087-43c619f2-fd3a-11e7-8cb4-bb0994cc3fa9.png">

## What were you trying the project on ?
  - Browser: Chrome
  - Operating System: OS X

Issue #214, opened 2018-01-19
source showing as FALSE in search


ManjotS:
## What went wrong ?
source showing as false for second tuk here: https://www.sikhitothemax.org/search?q=sBQ&type=0&source=all

## What should have happened instead ?
show nothing instead of false

## How can authors of this project cause the issue to occur on their systems ?
https://www.sikhitothemax.org/search?q=sBQ&type=0&source=all

PR #236, opened 2018-01-26
fix-source-false

tarunsingh5:
fix raag source showing as 'false' for dasam bani

### DIFF
--- a/src/js/search.js
+++ b/src/js/search.js
@@ -66,13 +66,12 @@ function loadResults(offset) {
   });
 }
 
-function getShabadHyperLink (shabad) {
+function getShabadHyperLink(shabad) {
   return `/shabad?id=${shabad.shabadid}&q=${q}${type ? `&type=${type}` : ''}${source ? `&source=${source}` : ''}`;
 }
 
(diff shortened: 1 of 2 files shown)


tarunsingh5:
#216 #214 fixed by this

src/js/search.js
```diff
@@ -70,6 +70,11 @@ function getShabadHyperLink (shabad) {
   return `/shabad?id=${shabad.shabadid}&q=${q}${type ? `&type=${type}` : ''}${source ? `&source=${source}` : ''}`;
 }
 
+function getRaagOrNull(Raag) {
+  if(Raag == "No Raag" ||  Raag == null){ return ""}
```
bogas04: Could we please have it as `raag` since it's not a class per-se?

bogas04: Also, there are some linting errors. Try to run `npm run test`. To find those.

tarunsingh5:
i changed Raag to raag and removed linting errors from my changes

navdeepsinghkhalsa:
Fixes implemented

PR #255, opened 2018-01-28
padched copy fix

tarunsingh5:
removed padding for the akhan class and instead using a span to insert a nbsp for padched

### DIFF
--- a/src/js/renderShabad.js
+++ b/src/js/renderShabad.js
@@ -22,7 +22,7 @@ function renderShabad(gurbani, nav) {
     .concat(prefs.shabadToggles)
     .forEach(option => document.getElementById(option).click());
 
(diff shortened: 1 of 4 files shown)


src/js/renderShabad.js
```diff
@@ -22,7 +22,10 @@ function renderShabad(gurbani, nav) {
     .concat(prefs.shabadToggles)
     .forEach(option => document.getElementById(option).click());
 
-  Object
+    var larivaarStatus = localStorage.getItem('shabadToggles').indexOf('larivaar-toggle')>-1;
```
bogas04: Is `shabadToggles` a preference? In that case we should use `getPref` method to avoid conflicts in future.

PR #271, opened 2018-01-29
Use fetch instead of jQuery

bogas04:
We are getting very close to removing jQuery! 💃

### DIFF
--- a/src/js/search.js
+++ b/src/js/search.js
@@ -1,8 +1,16 @@
-function loadResults(offset = null) {
+const $searchResults = document.querySelector('.search-results');
+
+function loadResults({ source, type, q, offset = null }) {
+
   return fetch(Khajana.buildApiUrl({ q, type, source, offset }))
     .then(r => r.json())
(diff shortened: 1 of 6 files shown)


src/js/search.js
```diff
@@ -84,43 +66,68 @@ function getShabadHyperLink(shabad) {
...
+
+  [...$searchResults.querySelectorAll('a.load')]
+    .forEach(el => el && el.addEventListener('click', () => loadResults(el.dataset.nextpage)));
+
+  if (q === '') {
+    $searchResults.appendChild(<h3><span>Please enter your query in the search bar above</span></h3>);
+    return;
+  }
+
+  document.body.classList.toggle("loading");
```
navdeepsinghkhalsa: Could you please use single quotes here?

bogas04:
Removed another piece of jQuery (not quite sure about the remove logic I wrote). I guess we can officially drop jQuery from dependencies now! 👯Next task would be to use no global variables and then finally port to React. I know this seems to be a very slow way of achieving it, but this ensures that feature addition isn't blocked coz of architectural changes.

PR #278, opened 2018-01-30
Hide meta data borders when empty

bogas04:
Hides borders of meta data area when it's empty. You can notice these borders during loading states.

* [x]  Ran `npm test`.
* [x]  Searched a shabad successfully.


ManjotS:
lgtm

PR #294, opened 2018-02-09
Use React and Webpack

bogas04:
Mega PR that re-writes application in ReactJS.

## Organizational Changes

* [x] File re-org
  * [x] `pages/` & `components/`
* [x] New Dependencies
  * [x] Add React to dependencies.
  * [x] Use babel@7.
  * [x] Redux.
* [x] Polyfills
  * [x] fetch (modern ajax replacement)
  * [x] Object.entries (convenience)
  * [x] Array.prototype.contains (convenience)
  * [x] String.prototype.includes (convenience)
  * [x] Map (React 16 requires this)
  * [x] Set (React 16 requires this)
  * [x] requestAnimationFrame (React 16 & our code-base requires this)

## Functional Changes

* [x] Persist redux store.
* [x] Port core components to React
  * [x] Gurmukhi Keyboard.
  * [x] SearchForm.
  * [x] SearchForm for non-home pages.
  * [x] Display Options. (90% done)
  * [x] Shabad (95% done, need to fix copy click & footer)
  * [x] Font Size doesn't update.
  * [x] Loading Component.
  * [x] Larivaar Assist.
* [x] Port pages
  * [x] Home (Fully working)
  * [x] About
  * [x] Terms of Service
  * [x] Help
  * [x] 404
  * [x] Search (need to wire in Load More)
  * [x] Ang
  * [x] Shabad Page
  * [x] Hukamnama
  * [x] Random

## Regressions
* [ ] Placeholder animation breaks when it's changed while an animation is in progress.
> Not sure how to fix this in a reactive way. For now `<select />` is disabled during animation.
* [x] Copy to clipboard for baani lines doesn't work.
> This can be fixed in a better way without relying on DOM.
* [ ] Load More doesn't append results to current list, but loads fresh list.
> I deliberately did this, now we can have shareable offset-ed links without having to worry about loading & scrolling to said place.
* [ ] Some animations are lost.
> This has happened coz now instead of hiding elements from CSS we just remove from DOM. Can be done based on severity of those elements.
* [ ] Font size steps are bit different now.
> Let me know if you want it in a different way.
* [ ] Previous preferences aren't respected.
> I used new localstorage keys, can be updated if required.

## Optimizations
* [x] Use PureComponents over functional components.
> Great with new class fields 💃 
* [x] Add prop-types to all
> Anyone wants to take this up?
* [x] [Remove](https://www.npmjs.com/package/babel-plugin-transform-react-remove-prop-types) prop-types for prod build
* [x] Use async routes (code-splitting).
> Now that {about,tos,help} are in-memory, it's really bad for users who might not even visit it. We need to split route based at least.
* [x] Write basic tests in jest.
> I'll write some to get us started.
* [x] Optimize production build.
> Webpack configs.
* [ ] Use server side rendering.
> This will help us a lot with SEO and perf.

## Final Tests:


bogas04:
As you can see, we are almost done restoring functionality with these changes.

I'll take a break now (need to focus on day job). Will take up remaining parts tomorrow.

PR #298, opened 2018-02-10
Fixed analytics test to use .org instead of .com

ManjotS:


tarunsingh5:
lgtm

PR #321, opened 2018-03-02
Fix #70

### LINKED ISSUE #70: Use React and Webpack
Tracking progress made by https://github.com/KhalisFoundation/SikhiToTheMax-Web/pull/294 here.

## Regressions
* [ ] Placeholder animation breaks when it's changed while an animation is in progress.
  Search mode drop down is gray until animation completes
> Not sure how to fix this in a reactive way. For now `<select />` is disabled during animation.
* [ ] Some animations are lost.
> This has happened coz now instead of hiding elements from CSS we just remove from DOM. Can be done based on severity of those elements.
* [ ] Previous preferences aren't respected.
> I used new localStorage keys, can be updated if required.

## Cleanup
* [ ] Clean up old localStorage keys.

## Fixed regressions
* [x] Copy to clipboard for baani lines doesn't work.
> This can be fixed in a better way without relying on DOM.
* [x] Load More doesn't append results to current list, but loads fresh list.
> I deliberately did this, now we can have shareable offset-ed links without having to worry about loading & scrolling to said place.
* [x] Font size steps are bit different now.
> Let me know if you want it in a different way.
issues found:
* [x] All share links are naked domain (not currently URI).
* [x] Gurmukhi margins don't match non react - Gurmukhi is too close to line above and too far from line below.
* [x] Transliterations don't show by default and the button does nothing.
* [x] Default load of search or shabad has font size max, larivaar and larivaar asisst on - should be no larivaar and size is `1.6em` (`16` on slider).
* [x] Display and font options don't float (flex) - they are using mobile style of one option/title per line. they should flex to to the right when not mobile and then go down (title above option, see non react).
* [x] Right drop down defaults to Guru Granth Sahib Jee - should be all
* [x] Pages are 100% width instead of current width that pads page
* [x] `package.json` previously replaced `api.shabados.org` with `devapi.khajana.org` (ONLY FOR DEV), this is gone and means we cant troubleshoot p0 issues with api - additionally, make it future proof and have it replace api.banidb.com.
> Done at Webpack build time. Also updated shabados library to accept API_URL. Do I need to rename it to banidb now? 😄 

## Optimizations
* Use server side rendering. Tracked in #329
> This will help us a lot with SEO and perf.
*  Figure out common components in route chunks and move them to `app.js`. `<ShabadContent />` for example. Tracked in #330
* [x] Add `<link preload />` for all important routes for predictive fetching.
  * [x] `/home` preloads `/search`.
  * [x] `/search` preloads `/ang` & `/shabad`.
  * [x] If metrics show some other route is very popular we can preload it as well.
* [x] Use a real router to avoid actually loading URLs.

bogas04:
Mega PR to fix all issues in #70 


bogas04:
Need to leverage react-router completely next, then we can go for server side rendering

* [ ] Use `<Link /`>
* [ ] Use react-router redirect instead of `{replaceState, pushState, redirectTo}`.
* [ ] Update `routes.js` to detect if it's in server env, and use "require" in that case instead of dynamic imports.
* [ ] Use `react-dom/server#renderToNodeStream`

PR #385, opened 2018-04-03
Fix for iPhone

bogas04:
Also upgraded the dependencies


bogas04:
@navdeepsinghkhalsa for some reason the css isn't updated

navdeepsinghkhalsa:
bundle.css has a bogus version string that is a relic from the PHP site. It could be creating a cache issue

Issue #329, opened 2018-03-16
Add server rendering


bogas04:
Now that we're on react and have isomorphic rendering code, we can go ahead and server render our pages. We'll get a lot of benefits like:

* Faster meaningful paint.
* Stream HTML.
* Add better meta tags and SEO related details. #317 

How to do it?

* [ ] Add `node-fetch` to load details on server end, this is required for SEO and rendering content on search/shabad pages.
* [x] Use templating, `marinate` probably to generate different HTMLs per route.
* [ ] Figure out server rendering (use `require()` in-place of `import()`)

saintsoldierx:
@bogas04 is this completed?

saintsoldierx:
Navdeep, Gurpreet - what's your thoughts on this ticket. 2-3 years old now, is this still valid?

gsin11:
I can check this out, idea is good and definitely in favour of better SEO performance.

navdeepsingh:
@saintsoldierx For SSR purpose (https://github.com/KhalisFoundation/sttm-web/issues/375#issuecomment-641609314) @plug-n-play already working on it. We need to chalk out beforehand rather we opt for the Nextjs method or manually using marinate, webpack, etc.
cc @gsin11

navdeepsingh:
As this confirmed, we need to work towards: "it's better we add webpack+babel tooling for our server to be able to render react components". @gsin11 we can sync together to make it possible.

saintsoldierx:
@gurjit03 i want you take this up soon as well please. Get with Navdeep/Gurpreet to learn more about it.

gurjit03:
@saintsoldierx we can close this as our direction is more towards next.js now.

PR #431, opened 2018-04-25
Use marinate as template, add basic SEO

### LINKED ISSUE #318: SEF URLs
## What went wrong ?
Dynamic URL's are currently being generated when searching. Having static, meaningful URL's will help our SEO.

bogas04:
* We use [esm](https://github.com/standard-things/esm) to add support for import/export syntax to nodejs.
* [marinate](https://github.com/bogas04/marinate) is used to compose the template. It's in-house thingy with easy ways of dealing with streams and async data. Will come handy when we do React SSR streaming and prefetch data at server. We can also stream the CSS inlined instead of 2 separate `<link />`s. But before that I guess we should get rid of foundation. 

* This opens up the gates for SSR. Next thing we need to do is make server learn JSX (webpack), so that we can use it to render React app at server side. Would be taken care by #329. Once we do that, it'll fix #318 and will also make our app load much faster.


bogas04:
@tarunsingh5 I want you to review the SEO title and descriptions.

PR #440, opened 2018-04-29
New Features!

bogas04:
# :racing_car: Performance:
 * Migrated to Webpack 4 thanks to @Harpreet Singh Jee’s contributions!
 * Because of above change, we have reduced our overall JS footprint
    * by 26% (For: landed to home > searched > opened a result).
    * by 2% (landed to home but bounced back).
 * NodeJS updated to 8 by @navdeepsingh, which is 30-50% faster compared to 7, and supports newer syntaxes better, this means better server response times.

# :clap: User Interface & Visibility:
 * We now set a cookie for darkmode, so that initial load does do a splash of white.
 * theme-color meta info now makes chrome UI blue 💃  
 * Grey out next/previous buttons instead of hiding them in Shabad/Ang view.
 * Basic title/description for SEO, this will increase our organic traffic somewhat.

# :wrench: Fixed:
 * Placeholder animation doesn’t disable search type dropdown
 * Search type/source/query now match URL set fields if present, else whatever you selected last time.
 * Backspace of Gurmuhki Keyboard doesn’t add undefined text to query
 * Share links aren’t naked domain links anymore

# :new:  Added:
 * Auto redirect HTTP to HTTPS.
 * ~~Service worker added for offline support and faster repeat loads.~~ To be added in next release after proper testing.
 * Copy all button.
 * Copy short link input.
 * Embed feature.


embed/script.js
```diff
@@ -0,0 +1,252 @@
+function h() {
```
jhbsk: A more descriptive function name would be good - as it's not obvious from the outset

embed/script.js
```diff
@@ -0,0 +1,252 @@
...
+  var attributes = arguments[1] || {};
+  var children = Array.prototype.slice.call(arguments, 2);
+  var el = document.createElement(type);
+
+  if (children[0] instanceof Array) {
+    children = children[0];
+  }
+
+  Object.keys(attributes)
+    .map(function(key) {
```
jhbsk: Are we using the new fat arrow syntax: `(key) => { }` or old school style: `function() { }`

We should just stick with one of these styles (I would suggest the first one) as it greatly improves readability

jhbsk: I'm seeing both styles across the codebase - so maybe a find & replace is in order!

embed/script.js
```diff
@@ -0,0 +1,252 @@
...
+            value(e);
+          },
+          false
+        );
+      } else {
+        el.setAttribute(key, value);
+      }
+    });
+
+  if (children instanceof Array) {
```
jhbsk: Just to smooth things out, we should maybe have our own type checking function e.g. https://gist.github.com/jhbsk/11083919

Maybe one for the future!

embed/script.js
bogas04: Support for fat arrow syntax is still poor so I need to transpile it. We want to keep the code in `embed.js` loadable by any/most browser.

embed/script.js
```diff
@@ -0,0 +1,252 @@
...
+      }
+    });
+}
+
+(function() {
+  function addControls() {
+    const $root = document.querySelector('#sttm-root');
+    const $controls = $root.querySelector('.sttm-shabad-controls');
+
+    const $transliteration = Array.prototype.slice.call(
```
jhbsk: This whole block so ideally be wrapped in a:

```
if ($root) {
    // ...
}
```

embed/script.js
bogas04: Vahiguru Jee. `h` is actually the name of this lib, inspired from [hyperscript](https://github.com/hyperhype/hyperscript). We used to use this when our codebase was in jQuery, and sttm-desktop still uses it. So the devs are versed with its functionality. Either way, we can refactor this script anytime in future.

seo/index.js
```diff
@@ -0,0 +1,58 @@
+import { SOURCES } from '../src/js/constants';
+
```
jhbsk: Nice!

embed/script.js
bogas04: Hanjee. Since this script would be loaded in external pages, idea is to keep it as lightweight as possible.

embed/script.js
bogas04: Ideally yes, however `$root` is totally owned by us.

embed/script.js
bogas04: Moved all to `function`s.

jhbsk:
Looking really good overall ! The performance improvements are great

Issue #365, opened 2018-03-27
In case of an error, no links work.


bogas04:
## What went wrong ?

If something breaks, no links work.

## What should have happened instead ?

Links should work.

## How can authors of this project cause the issue to occur on their systems ?

  1. Trigger an error (Add react dev tool ext to chrome, search for "Fetch", change error: null to "true").
  2. Click any link (logo/hukamnama etc.)

## What were you trying the project on ?

  - Browser: Chrome
  - Browser Version: 66
  - Operating System: OSX

rajkambo:
After digging into the problem with the GenericError being rendered in Root.js in #473 it looks like we need to redesign how errors are handled in the app. The problem really starts with this attempt at a "catch all" GenericError: https://github.com/KhalisFoundation/sttm-web/blob/master/src/js/Root.js#L29. What happens is the error page gets rendered correctly, but the Routes don't get set up. So after rendering the error, any subsequent clicks don't do anything. 

I read this great article: https://reactjs.org/blog/2017/07/26/error-handling-in-react-16.html, and took a similar approach in #473. In this case, any error related to searching never propagates up to the Root.js so all the Routes remain in place after the GenericError has been rendered.

So IMO what we need to do is create a generic "HandleGenericError" component, and have it wrap everything that gets returned in the render calls of all the route specific page's index.js's (e.g. what was done in the Search/index.js here: https://github.com/KhalisFoundation/sttm-web/blob/47435e056677933dde9afb170c5b2cb5abb303e8/src/js/pages/Search/index.js#L39)

bogas04:
@rajkambo I love your debugging style.

Sorry I couldn't help you out before. So problem is that we are catching errors at a level above Router, and `<Links />` have no idea what to do without `<Router />`. Secondly, we always render error based on state, but never update it.


I've moved the `componentDidCatch` at Layout level, coz that's something we always render for each route.

rajkambo:
@bogas04 looks great! thanks for the detail! 👍 And no problem :D So long as the team gets the job done, it's all good :D

PR #473, opened 2018-05-25
348 add timeout to fetch requests

rajkambo:


src/js/components/Fetch.js
```diff
@@ -1,6 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
+const TIMEOUT = 10000; // 10 seconds
```
bogas04: Could you refactor it so that Fetch accepts timeout as a prop and defaults to 10s?

rajkambo: Yup, for sure! 👍

rajkambo:
So I updated the PR with timeout getting passed as a prop, but I found that timing out seems to cause #365:
image

Will do some more digging into how we can fix this 👍

bogas04:
@rajkambo were you able to fix the bug?

rajkambo:
@bogas04 just mulling over what would be the best way to handle the issue. So the root cause of this error and #365 is the same -- a trying to access fields from a the null "data" object here: https://github.com/KhalisFoundation/sttm-web/blob/master/src/js/pages/Search/index.js#L44. 

This is caused by the error being null here: https://github.com/KhalisFoundation/sttm-web/blob/master/src/js/pages/PageLoader.js#L17. 

Since error the error is null here data (null) is passed on to the children and a null exception occurs leading to the generic: https://github.com/KhalisFoundation/sttm-web/blob/master/src/js/Root.js#L29.

In my specific case (on timed out Fetch requests), I could modify the setTimeout call to have a third argument and make it something like "Fetch request timed out", this would then ensure that error isn't null as it is passed on here: https://github.com/KhalisFoundation/sttm-web/pull/473/files#diff-10ef1ed613a66115bbbefad359a37a6cR67.

In this case we end up back at: https://github.com/KhalisFoundation/sttm-web/blob/master/src/js/pages/PageLoader.js#L17. Where an error is thrown when "error" isn't null. And since this error doesn't get caught anywhere we end up right here: https://github.com/KhalisFoundation/sttm-web/blob/master/src/js/Root.js#L29.

And in since the error goes uncaught, we end up in a state where we can't click anything.. 

Going to do some more reading on how exactly error handling works in React, but would be great if you had any suggestions :)

rajkambo:
Update: I think I got the ideal fix! Will commit soon 👍

src/js/components/HandleSearchError.js
```diff
@@ -0,0 +1,58 @@
...
+import { ACTIONS, errorEvent } from '../util/analytics';
+import GenericError, { SachKaur } from './GenericError';
+import { TYPES, SOURCES, PLACEHOLDERS, TEXTS } from '../constants';
+
+export default class HandleSearchError extends React.PureComponent {
+  static propTypes = {
+    q: PropTypes.string.isRequired,
+    type: PropTypes.number.isRequired,
+    source: PropTypes.string.isRequired,
+    children: PropTypes.node.isRequired,
```
rajkambo: Wasn't sure about the type of "children" but read this: https://stackoverflow.com/questions/42122522/reactjs-what-should-the-proptypes-be-for-this-props-children and settled on "node".

src/js/components/HandleSearchError.js
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
```
rajkambo: Created this component for the error handling approach suggested here: https://reactjs.org/blog/2017/07/26/error-handling-in-react-16.html.

rajkambo:
@bogas04 took a while to dig into this issue, and I fixed this specific error case, but it looks like we need to fix error handling as a whole in the project (as the catch all we had in Root.js isn't going to cut it). I left a detailed comment in #365  on why the links aren't working and what I think we need to do to reform our error handling👍

rajkambo:
@bogas04 could I get the thumbs up when you get the time 👍 Would be great to wrap this one up :D

bogas04:
@rajkambo do we need HandleSearchError or GenericError catches the fetch timeout fine?

bogas04:
I added the logic at our single ErrorBoundary (Layout) to keep logic simple.

@rajkambo thank you so much!

screen shot 2018-06-15 at 7 42 04 pm

PR #495, opened 2018-06-08
added charset.. 

ManjotS:
I am unsure if text/html is right or application/javascript or application/json

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -20,7 +20,6 @@ app
 
   // Infrastructure display
   .use((req, res, next) => {
-    res.setHeader('Content-Type', 'text/html; charset=utf-8');
     res.setHeader('origin-server', hostname);
     return next();
   })
@@ -30,6 +29,7 @@ app
 
   // Direct all calls to index template
   .get('*', (req, res) => {
+    res.setHeader('Content-Type', 'text/html; charset=utf-8');
     const { path } = req;
 
     const { title: _title, createDescription } = seo[
--- a/server/index.js
+++ b/server/index.js
@@ -20,7 +20,7 @@ app
 
   // Infrastructure display
   .use((req, res, next) => {
-    res.setHeader('Content-Type', 'text/html');
+    res.setHeader('Content-Type', 'text/html; charset=utf-8');
     res.setHeader('origin-server', hostname);
     return next();
   })


rajkambo:
Hm, doesn't it depend on what the client is looking for though? Would be great to have more context :D

ManjotS:
The main issue I am trying to fix is cdn (https://sikhitothemax.azureedge.net) which is trying to force download of the home page.

bogas04:
@rajkambo currently our server is designed in a way wherein it serves static assets from filesystem first and when nothing matches, it serves the website, that handles the route or 404 page.

rajkambo:
Ah I see. That's some interesting default behavior... But yes, setting the content-type to text/html then makes perfect sense if we're trying to serve to the browser 👍

bogas04:
@rajkambo hanjee. And express will infer content type from extension for static files automatically.

ManjotS:
deployed to dev but it broke dev

bogas04:
Hmm, was it working locally?

rajkambo:
Weird I checked out http://dev.sikhitothemax.org/ no response, but got a 200 back..

image

rajkambo:
Weird, a simple `curl https://dev.sikhitothemax.org` gives the correct response (same as `curl https://www.sikhitothemax.org`)

ManjotS:
do `curl -i` which gives headers

ManjotS:
https://blog.stackpath.com/accept-encoding-vary-important

rajkambo:
I'll give that a read 👍 Response is the same for both, but the headers differ:

DEV:
```
HTTP/1.1 200 OK
Server: nginx/1.12.1 (Ubuntu)
Date: Fri, 08 Jun 2018 05:14:39 GMT
Content-Type: text/html; charset=utf-8
Transfer-Encoding: chunked
Connection: keep-alive
X-Powered-By: Express
origin-server: use
Vary: Accept-Encoding

```
PROD:
```
HTTP/1.1 200 OK
Server: nginx/1.12.1 (Ubuntu)
Date: Fri, 08 Jun 2018 05:15:30 GMT
Transfer-Encoding: chunked
Connection: keep-alive
X-Powered-By: Express
origin-server: use

```

ManjotS:
I will let you experts sort this out. 

Basically, need cdn to stop thinking home page is a download.

rajkambo:
Going to dig into this more tomorrow, but one last thing. I noticed it looks like response (from dev) is compressed (Content-Encoding: gzip). I was catching in in the #sttm-web channel and it looks like the npm compression library was modified by Jujhar recently (he ran `npm install compression`). Perhaps its related? 

image

ManjotS:
curl headers from CDN (which is pulling from PROD):
```
HTTP/2 200 
accept-ranges: bytes
content-type: application/octet-stream
date: Fri, 08 Jun 2018 05:52:49 GMT
last-modified: Fri, 08 Jun 2018 01:59:11 GMT
origin-server: use
server: ECAcc (saa/83DD)
x-cache: HIT
x-powered-by: Express
content-length: 4733
```

ManjotS:
@rajkambo he did that on the old server. Current servers were built by me and are brand new as of 2 days ago.

rajkambo:
Oh weird, I usually see octet streams used in conjunction with multipart requests for streaming stuff. What was the url that you curled to get that? 

So it looks like the CDN streams instead of statically serving?

ManjotS:
CDN: sikhitothemax.azureedge.net

It is defaulting to that because it isn't getting a content type from nodejs.

ManjotS:
CDN header for api.banidb.com:
```HTTP/2 200 
accept-ranges: bytes
access-control-allow-origin: *
content-type: application/json; charset=utf-8
date: Fri, 08 Jun 2018 05:59:47 GMT
last-modified: Thu, 07 Jun 2018 15:58:02 GMT
server: ECAcc (saa/83DD)
vary: Accept-Encoding
x-cache: HIT
content-length: 134```

navdeepsinghkhalsa:
@ManjotS Could you please try `res.contentType('text/html');`?

ManjotS:
So that renders the same header, so this is a better way to do it.

BUT, same issues. I also tried application/javascript

The site doesn't load.

PR #497, opened 2018-06-08
trying application/json content type as used by api

ManjotS:


rajkambo:
What's up with the package-lock?

PR #501, opened 2018-06-11
Add reset to font and display options

rajkambo:

### DIFF
--- a/src/js/components/Controls.js
+++ b/src/js/components/Controls.js
@@ -3,11 +3,11 @@ import { connect } from 'react-redux';
 import ShabadControls from './ShabadControls';
 import ShareButtons, { supportedMedia as _s } from './ShareButtons';
 import {
-  setUnicode,
   setFontSize,
   setTranslationLanguages,
   setTransliterationLanguages,
-  setDarkMode,
+  resetDisplayOptionsDispatcher,
+  resetFontOptionsDispatcher,
   toggleDisplayOptions,
   toggleFontOptions,
   toggleLarivaarAssistOption,
@@ -18,6 +18,7 @@ import {
   toggleSplitViewOption,
   toggleDarkMode,
 } from '../features/actions';
+import store from '../features/store'
 
 export const supportedMedia = _s;
 
@@ -69,12 +70,15 @@ class Controls extends React.PureComponent {
 // TODO: Take exactly what we need.
 const stateToProps = state => state;
 
+const resetFontOptions = resetFontOptionsDispatcher(store.dispatch)
+const resetDisplayOptions = resetDisplayOptionsDispatcher(store.dispatch)
+
 const dispatchToProps = {
-  setUnicode,
   setFontSize,
   setTranslationLanguages,
   setTransliterationLanguages,
-  setDarkMode,
+  resetDisplayOptions,
+  resetFontOptions,
   toggleDisplayOptions,
   toggleFontOptions,
   toggleLarivaarAssistOption,
--- a/src/js/components/Controls.js
+++ b/src/js/components/Controls.js
@@ -6,8 +6,8 @@ import {
   setFontSize,
   setTranslationLanguages,
   setTransliterationLanguages,
-  resetDisplayOptionsDispatcher,
-  resetFontOptionsDispatcher,
+  resetDisplayOptions,
+  resetFontOptions,
   toggleDisplayOptions,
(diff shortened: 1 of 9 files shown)


rajkambo:
So unfortunately with the addition of this feature, I wasn't able to keep the same exact "smooth" scrolling when your changing the font-size :( 

Right now, when you try to change the font-size, you can drag the bar left and right. But with the addition of this, the user will have to click on a spot on the line change the font-size. 

The reason for this is that we have a new way of changing the font-size now (when you click reset), and thus, have to listen for these changes and apply these changes by setting the "value" of the input. Previously we didn't have to worry about this because the only way of changing the value of the input was by sliding the bar left and right so we never had to explicitly set the value of the input type.

TLDR; changing the font-size still works, but we don't have the same smooth scrolling animation. Any suggestions are welcome if I've overlooked something :)

bogas04:
Great work!

Could you move the setters into a "compound" action that resets display and font options?

bogas04:
@rajkambo Could you please take some time to do the same?

What i'm asking for is;

```js
const resetDisplayOptions = payload => dispatch => {
  dispatch(setA(DEFAULT_A));
  dispatch(setB(DEFAULT_B));
  dispatch(setC(DEFAULT_C));
}

const resetFontOptions = payload => dispatch => {
  dispatch(setP(DEFAULT_P));
  dispatch(setQ(DEFAULT_Q));
}
```

rajkambo:
Thanks for the feedback :D Sorry for the delay, was away from my home PC this week. I took a look and the latest commit I pushed was my first attempt at compound actions. The problem with this solution is two-fold, however:

(1) The passing of the dispatcher to the resetDisplayOptions and resetFontOptions within Controls.js: https://github.com/KhalisFoundation/sttm-web/pull/501/files#diff-191162c47985f57e4d0da46b7f93272eR73 doesn't seem very elegent. Is there a better way you think we could handle this?

(2) I noticed in console, I now see a: TypeError: action is undefined, whenever I click a reset button. The reset still works but I'd like to get to the bottom of why this occurs :)

src/js/features/actions/index.js
```diff
@@ -50,3 +65,17 @@ export const SET_TRANSLITERATION_LANGUAGES = 'SET_TRANSLITERATION_LANGUAGES';
...
+export const SET_DARK_MODE = 'SET_DARK_MODE';
+export const setDarkMode = createAction(SET_DARK_MODE);
+
+export const resetDisplayOptionsDispatcher = dispatch => () => {
+  dispatch(setTransliterationLanguages(DEFAULT_TRANSLITERATION_LANGUAGES));
+  dispatch(setTranslationLanguages(DEFAULT_TRANSLATION_LANGUAGES));
+  dispatch(setDarkMode(DEFAULT_DARK_MODE));
+}
+
+export const resetFontOptionsDispatcher = dispatch => () => {
```
bogas04: You need to accept payload first and then (dispatch, getState). For that to work we'll need [redux-thunk](https://github.com/reduxjs/redux-thunk/blob/master/README.md). Sorry I forgot to mention that earlier.

src/js/components/Controls.js
```diff
@@ -67,10 +70,15 @@ class Controls extends React.PureComponent {
 // TODO: Take exactly what we need.
 const stateToProps = state => state;
 
+const resetFontOptions = resetFontOptionsDispatcher(store.dispatch)
+const resetDisplayOptions = resetDisplayOptionsDispatcher(store.dispatch)
+
```
bogas04: You don't need to do this if you add redux-thunk middleware.

src/js/components/ShabadControls.js
```diff
@@ -190,12 +205,20 @@ export default class ShabadControls extends React.PureComponent {
...
                 onInput={e => setFontSize(e.currentTarget.value / 10)}
               />
               <big className="gurbani-font">A</big>
             </div>
+            <div className="display-reset-option-content">
+              <a
+                className={`display-option-toggle ${darkMode ? 'active' : ''}`}
+                onClick={resetFontOptions}
+              >
+                Reset
```
bogas04: Could you please use TEXTS from constants file? It'll help us in internationalizing later. `TEXTS.RESET`

rajkambo:
Added constants for all the string literals I saw in that file, let me know if I missed anything :D

PR #587, opened 2018-12-16
Add Service Worker Support Fix #588

### LINKED ISSUE #68: Adapt to be a PWA
[https://developers.google.com/web/progressive-web-apps/checklist](https://developers.google.com/web/progressive-web-apps/checklist)

**Baseline Progressive Web App Checklist**
- [x] Site is served over HTTPS
- [x] Pages are responsive on tablets & mobile devices
- [x] The start URL (at least) loads while offline (Service worker)
- [x] Metadata provided for Add to Home screen
- [x] First load fast even on 3G (Code splitting, SVGs)
- [x] Site works cross-browser
- [x] Page transitions don't feel like they block on the network (React-router)
- [x] Each page has a URL (We're an SPA with proper routes)

### LINKED ISSUE #588: Service worker support
Final step stopping sttm from becoming a full PWA.

Why?

* Website loads even when there's no network. We can tweak it to load whatever SG/Angs the user has ever loaded.
* Repeat visits are much much faster.

Why not?

* Doing it incorrectly can lock users on an old version of website with only way to clear being to delete all site storage settings.

bogas04:
Fix #68/#588 by adding SW and offline support.

We need to have a kill switch so as to unregister all our user's SW if it's faulty.


bogas04:
I'm reading through workbox, seems quite simple in approach.

https://developers.google.com/web/tools/workbox/guides/precache-files/webpack

bogas04:
@navdeepsinghkhalsa kindly take out time to review this, I just want to make sure this would never cause a case where users are locked out in an old version of website.

Issue #58, opened 2017-01-25
Ability to choose fonts


irvanjitsingh:
Ability to change the default ASCII/unicode fonts

bogas04:
Is this resolved ?

ManjotS:
no.. this would fall under display options

bogas04:
What exactly are we looking at here?

ManjotS:
similar to font chooser in sundar gutka options.

We can use those fonts in fact (web akhar +thick +heavy, anmol lipi, etc)

ManjotS:
most likely just a drop down in font options

tarunsingh5:
fixed

PR #665, opened 2019-05-06
Ability to change fonts

Gauravjeetsingh:

Fixes Issue #58 

Added a dropdown to change fonts. If we enable the Unicode mode, the font family dropdown will disappear. Since there will be no change with changing the font family when Unicode is enabled.

8e8ih-5630h

With this, we got a small CSS issue only in google chrome. The issue is that when we open the font family dropdown, our page scrolls to top.

This is because the select dropdown acts weird in a position sticky container. I will be working on implementing similar behavior of our container without a sticky position.

Until I fix that, if there's any other feedback on the current implementation of changing font family, let me know.


Gauravjeetsingh:
Instead of a separate toggle, added Unicode as an option in the font dropdown
Screenshot from 2019-05-25 17-57-39

Issue #122, opened 2017-03-16
Add option to display options to remove ALL sharing links


ManjotS:

bogas04:
Do we need this as we hide them for mobile and show those only on hover for desktop?

ManjotS:
prefer the option

bogas04:
Where should we keep this option then?

tarunsingh5:
i think you put the < logo and when ppl click that it pops down w/ the share options

ManjotS:
@tarunsingh5 this ticket is to remove all sharing links from the page as an option. As said in the title its a toggle in the display drop down.

bogas04:
Okay;

We show share button only when page is scrolled to top.

* On mobile without WebShare API, we show one sharing option; WhatsApp button
* On mobile with WebShare API, we show one sharing option; Share button
* On desktop without WebShare API, we show on sharing option; Embed
* On desktop with WebShare API, we show 2 sharing options; Embed and Share.
* On desktop upon hovering a line, we 1 sharing option, tweet button.

Apart from that, we've short URL and copy button.

Since the share button doesn't really come in the way, I am not sure if this will change anything as we still show Display Options, Font Options, Larivaar and Larivaar Assist option all the time.

I think, we can instead have a full screen option on mobile, that hides the sticky bar for better reading experience. They can access it again by scrolling to top, clicking "Display Options" and then disabling it. If we get enough traction via analytics, we can evolve the feature further.

What do you think? @saintsoldierx @tarunsingh5

saintsoldierx:
![image](https://user-images.githubusercontent.com/27791330/50064723-7f450600-0167-11e9-9bf2-2bc6241c4c9e.png)

@bogas04 i think this ticket is in reference to the sharing links line by line on a shabad as per the screenshot...

ManjotS:
yes this is just to hide the sharing link

bogas04:
Can we implement this in a better way instead? We can't access this on mobile, and on desktop it can be annoying.

saintsoldierx:
Refer to #668

PR #668, opened 2019-05-09
Shows sharing links on selecting text

Gauravjeetsingh:

I read the discussion in issue #122 and came with this solution.

Instead of showing the share actions on hover, it will now show when we select the text.
Even If we select a little part from the baani, it will select the entire line (including translation and transliteration) and shows it's share action links.

5notp-3uhwu

It also fixes the color change effect when we hover on the share action icon.


src/js/components/Baani.js
```diff
@@ -114,8 +114,32 @@ export default class Baani extends React.PureComponent {
...
+      const shareDiv = baaniDiv.querySelector('.share');
+      shareDiv.style.display = 'block';
+    }
+  };
+
   componentDidMount() {
     this._scrollToHiglight();
+    const lineDivs = document.querySelectorAll('.line');
+    const ms = this.makeSelection;
+    lineDivs.forEach(function(ld) {
```
bogas04: Can we instead do it in the React side? Like event listeners on the line components

Gauravjeetsingh: Fixed @bogas04

src/js/components/Baani.js
```diff
@@ -153,6 +173,8 @@ export default class Baani extends React.PureComponent {
             key={shabad.id}
             id={`line-${shabad.id}`}
             className="line"
+            onMouseUp={this.makeSelection}
+            onMouseDown={this.removeSelection}
```
bogas04: Wouldn't work for split view then right?

src/js/components/Baani.js
```diff
@@ -193,8 +215,6 @@ export default class Baani extends React.PureComponent {
               onFacebookClick={this.onFacebookClick(shabad)}
               onCopyClick={this.onCopyClick(shabad)}
             />
-
-            <br />
```
bogas04: I think this was added to keep new lines in the copy pasted text.

src/js/components/Baani.js
```diff
@@ -116,6 +116,26 @@ export default class Baani extends React.PureComponent {
     }
   };
 
+  makeSelection = e => {
+    const shareDivs = document.querySelectorAll('.share');
+    shareDivs.forEach(function(sd) {
+      sd.style.display = 'none';
+    });
```
bogas04: Use arrow function

src/js/components/Baani.js
```diff
@@ -116,6 +116,26 @@ export default class Baani extends React.PureComponent {
...
+    if (window.getSelection().toString()) {
+      const baaniDiv = e.currentTarget;
+      var range = document.createRange();
+      range.selectNode(baaniDiv);
+      window.getSelection().removeAllRanges();
+      window.getSelection().addRange(range);
+      const shareDiv = baaniDiv.querySelector('.share');
+      shareDiv.style.display = 'block';
+    }
```
bogas04: Let's make a helper function that takes a DOM node and selects it. That way the functionality and logic isn't coupled. Makes sense?

src/js/components/Baani.js
Gauravjeetsingh: I checked copy paste with and without br tag, it's same as following.

```
ਤੁਧ ਬਿਨੁ ਰੋਗੁ ਰਜਾਈਆਂ ਦਾ ਓਢਣੁ ਨਾਗ ਨਿਵਾਸਾਂ ਦੇ ਰਹਣਾ ॥
tudh bin rog rajaieeaa(n) dhaa oddan naag nivaasaa(n) dhe rahanaa ||
Without Thee, the taking over of quilt is like disease and living in the house is like living with serpents
```

src/js/components/Baani.js
Gauravjeetsingh: Yeah, it doesn't work in split view. Do we want it to work in split view?

src/js/components/Baani.js
Gauravjeetsingh: Fixed.

src/js/components/Baani.js
Gauravjeetsingh: Removed the function. Using a separate class to show.

src/js/components/Baani.js
bogas04: Let's keep this in helpers file so that other components can use as well. What do you think?

Gauravjeetsingh: Yeah, good idea. Done. :raised_hands:

PR #676, opened 2019-05-20
Added missing indexes

plug-n-play:
Added missing indexes based on http://old.sgpc.net/CDN/Siri%20Guru%20Granth%20Sahib%20in%20Gurmukhi,%20with%20Index%20%28Uni%29.pdf

### DIFF
--- a/src/js/pages/GranthIndex/raagIndices.js
+++ b/src/js/pages/GranthIndex/raagIndices.js
@@ -221,7 +221,43 @@ export default {
       },
       {
         name: 'Raag Jaijavanti',
-        pages: [1352, 1353],
+        pages: [1352, 1352],
+      },
+      {
+        name: 'Salok Sehaskriti Mahalla 1',
+        pages: [1353, 1353],
+      },
+      {
+        name: 'Salok Sehaskriti Mahalla 5',
+        pages: [1353, 1360],
+      },
+      {
+        name: 'Mahalla 5 Gaatha',
+        pages: [1360, 1361],
+      },
+      {
+        name: 'Funehe Mahalla 5',
+        pages: [1361, 1363],
+      },
+      {
+        name: 'Choubole Mahalla 5',
+        pages: [1363, 1364],
+      },
+      {
+        name: 'Salok Bhagat Kabir Jio Ke',
+        pages: [1364, 1377],
+      },
+      {
+        name: 'Salok Seikh Farid Ke',
+        pages: [1377, 1385],
+      },
+      {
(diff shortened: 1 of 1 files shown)


src/js/pages/GranthIndex/raagIndices.js
```diff
@@ -221,7 +221,43 @@ export default {
...
+      {
+        name: 'Salok Seikh Farid Ke',
+        pages: [1377, 1385],
+      },
+      {
+        name: 'Swayyee Sri Mukh Baak Mahalla 5',
+        pages: [1385, 1410],
+      },
+      {
+        name: 'Salok Vaaran & more',
```
bogas04: & more lol :D . Since all other strings are transliteration, let's keep this one like that only. `Salok Vaaran Te Vadheek`.

bogas04:
```suggestion
        name: 'Salok Vaaran Te Vadeek',
```

src/js/pages/GranthIndex/raagIndices.js
```diff
@@ -221,7 +221,43 @@ export default {
       },
       {
         name: 'Raag Jaijavanti',
-        pages: [1352, 1353],
+        pages: [1352, 1352],
+      },
+      {
+        name: 'Salok Sehaskriti Mahalla 1',
```
bogas04: Pronunciation/Transliteraion of Mahalla/Mehla is controversial, let's resort to M1/M5? Might have to change Salok M9 as well if we go that route.

@saintsoldierx what do you think?

src/js/pages/GranthIndex/raagIndices.js
```diff
@@ -221,7 +221,43 @@ export default {
...
+        pages: [1353, 1353],
+      },
+      {
+        name: 'Salok Sehaskriti Mahalla 5',
```
bogas04: @tarunsingh5 I remember some discussion about the transliteration for Sehaskriti. Any conclusion?

src/js/pages/GranthIndex/raagIndices.js
plug-n-play: (FYI - the transliteration Mahalla exists in other places on the page which I followed)

src/js/pages/GranthIndex/raagIndices.js
plug-n-play: Done

src/js/pages/GranthIndex/raagIndices.js
bogas04: Yeah that's why I think we might have to change it there as well.

PR #678, opened 2019-05-22
Kill switch for service worker

bogas04:


server/template.js
```diff
@@ -1,37 +1,84 @@
 import marinate from 'marinate';
 import { ONLINE_COLOR } from '../common/constants';
 
+const preloadImages = [
```
bogas04: Right now these are hard coded but eventually we would be reading these from a manifest file generated by webpack.

plug-n-play: Do we already have a issue created for the follow-up item we need to do?

server/template.js
```diff
@@ -1,37 +1,84 @@
...
+const preloadImagesHTML = preloadImages
+  .map(i => `<link async rel="preload" href="${i}" as="image">`)
+  .join('\n');
+
+const preloadFonts = [
+  '/assets/fonts/AnmolLipiSG.ttf?v=1',
+  '/assets/fonts/GurbaniAkharHeavyTrue.ttf?v=1',
+];
+const preloadFontsHTML = preloadFonts
+  .map(f => `<link async rel="preload" href="${f}" as="font"></link>`)
```
navdeepsinghkhalsa: Is there any reason this isn't self-closing?

bogas04: Nice catch, thank you!

PR #680, opened 2019-05-22
Use hashes in webpack JS assets

bogas04:


server/template.js
```diff
@@ -110,8 +136,14 @@ export default ({ url, bodyClass, title, description }) => marinate`
...
   <script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=fetch,Object.entries,Array.from,Array.prototype.includes,String.prototype.startsWith,String.prototype.includes,Map,Set,requestAnimationFrame,Array.prototype.@@iterator"></script>
 
+<<<<<<< HEAD
+  <script src="${manifest['vendor.js']}"></script>
+  <script src="${manifest['app.js']}"></script>
+</body>
+=======
   <!-- Source Code -->
   ${scriptsHTML}
+>>>>>>> 169583a2dc05864018ec6e2212109621b9ae8b25
```
bogas04: This doesn't seem to be right. Bad merge?

server/template.js
```diff
@@ -19,6 +28,23 @@ export default ({ url, bodyClass, title, description }) => marinate`
...
+  <link async rel="preload" href="/assets/fonts/GurbaniAkharHeavyTrue.ttf?v=1" as="font">
+  <link async rel="preload" href="/assets/fonts/AnmolLipiSG.ttf?v=1" as="font">
+
+  <!-- Preload JavaScript -->
+  <link async rel="preload" href="${manifest['Search.js']}" as="script">
+  <link async rel="preload" href="${manifest['Shabad.js']}" as="script">
+
+  <!-- Preload Images -->
+  <link async rel="preload" href="/assets/images/sttm_logo.png" as="image">
+  <link async rel="preload" href="/assets/images/logo-192x192.png" as="image">
```
bogas04: Bad merge? Dev has automated strings for this.

server/template.js
```diff
@@ -110,8 +136,14 @@ export default ({ url, bodyClass, title, description }) => marinate`
   <!-- Polyfills -->
   <script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=fetch,Object.entries,Array.from,Array.prototype.includes,String.prototype.startsWith,String.prototype.includes,Map,Set,requestAnimationFrame,Array.prototype.@@iterator"></script>
 
+<<<<<<< HEAD
+  <script src="${manifest['vendor.js']}"></script>
+  <script src="${manifest['app.js']}"></script>
+</body>
+=======
   <!-- Source Code -->
   ${scriptsHTML}
```
bogas04: `scriptsHTML` handles the script tags, you just need to update the array below.

server/template.js
```diff
@@ -110,8 +136,14 @@ export default ({ url, bodyClass, title, description }) => marinate`
   <!-- Polyfills -->
   <script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=fetch,Object.entries,Array.from,Array.prototype.includes,String.prototype.startsWith,String.prototype.includes,Map,Set,requestAnimationFrame,Array.prototype.@@iterator"></script>
 
+<<<<<<< HEAD
+  <script src="${manifest['vendor.js']}"></script>
+  <script src="${manifest['app.js']}"></script>
+</body>
+=======
```
bogas04: This isn't required as `scriptsHTML` takes care of this.

server/template.js
Gauravjeetsingh: Couldn't get you here. Which array are we talking about?

server/template.js
Gauravjeetsingh: I removed these lines, is there any other change required?

server/template.js
bogas04: Check at the bottom of the file and look for `scripts` array. 

https://github.com/KhalisFoundation/sttm-web/blob/dev/server/template.js#L173

server/template.js
bogas04: Remove all these lines as well. Check the *HTML variables, they're handling these now.

server/template.js
bogas04: <<<<< HEAD is part of git conflict. We need to resolve that. And the resolution is to remove the script part as `scriptHTML` is gonna put them.

server/template.js
Gauravjeetsingh: @bogas04 Can you check now, if all the required changes are done?

PR #683, opened 2019-05-25
Sttm-desktop sync with sttm-web

Gauravjeetsingh:
This exposes the existing sync page in sttm

### DIFF
--- a/src/js/components/HeaderMenu.js
+++ b/src/js/components/HeaderMenu.js
@@ -45,13 +45,14 @@ export default class Menu extends React.PureComponent {
               Index
             </Link>
           </li>
-          {/* // TODO Enable when desktop feature is ready
-          <li>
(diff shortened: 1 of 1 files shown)


src/js/components/HeaderMenu.js
```diff
@@ -45,13 +45,14 @@ export default class Menu extends React.PureComponent {
...
           </li>
-          {/* // TODO Enable when desktop feature is ready
-          <li>
-            <Link to="/sync" onClick={toggleMenu}>
-              Sync
-            </Link>
-          </li>
-          */}
+          {
+            /* // TODO Enable when desktop feature is ready */
```
bogas04: We can remove the todo

Gauravjeetsingh: I was actually waiting on the desktop feature to complete before removing the comment, exposed it so that sttm-desktop team can test while working on it.

PR #698, opened 2019-05-30
Reset Center Align & Reposition option

plug-n-play:

### DIFF
--- a/src/js/features/actions/index.js
+++ b/src/js/features/actions/index.js
@@ -5,6 +5,7 @@ import {
   DEFAULT_UNICODE,
   DEFAULT_FONT_SIZE,
   DEFAULT_FONT_FAMILY,
+  DEFAULT_CENTER_ALIGN_GURBANI,
 } from '../../constants';
 
 export const createAction = (type, meta) => payload => ({
@@ -71,6 +72,7 @@ export const resetDisplayOptions = () => dispatch => {
   dispatch(setTransliterationLanguages(DEFAULT_TRANSLITERATION_LANGUAGES));
   dispatch(setTranslationLanguages(DEFAULT_TRANSLATION_LANGUAGES));
   dispatch(setDarkMode(DEFAULT_DARK_MODE));
+  dispatch(setCenterAlignOption(DEFAULT_CENTER_ALIGN_GURBANI));
 };
 
 export const CHANGE_FONT = 'CHANGE_FONT';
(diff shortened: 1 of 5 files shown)


src/js/features/reducers/index.js
```diff
@@ -230,7 +232,22 @@ export default function reducer(state, action) {
...
+        centerAlignGurbani,
+      };
+    }
+    case SET_CENTER_ALIGN_OPTION: {
+      const centerAlignGurbani = action.payload;
+
+      saveToLocalStorage(
+        LOCAL_STORAGE_KEY_FOR_CENTER_ALIGN_VIEW,
+        centerAlignGurbani
+      );
```
bogas04: What's the difference between toggle and set?

plug-n-play: Toggle - when user toggles the option. Set, used when user clicks reset. Followed the same pattern as used in `TOGGLE_DARK_MODE` `SET_DARK_MODE`

tarunsingh5: @bogas04 is there any action needed here?

src/js/features/actions/index.js
```diff
@@ -7,6 +7,7 @@ import {
   DEFAULT_FONT_FAMILY,
   DEFAULT_CENTER_ALIGN_GURBANI,
 } from '../../constants';
+import store from '../../features/store';
```
bogas04: I think you can use `@/features/stores`

src/js/features/actions/index.js
```diff
@@ -87,5 +88,7 @@ export const resetFontOptions = () => dispatch => {
 export const SET_CENTER_ALIGN_OPTION = 'SET_CENTER_ALIGN_OPTION';
 export const setCenterAlignOption = createAction(SET_CENTER_ALIGN_OPTION);
 
-export const TOGGLE_CENTER_ALIGN_OPTION = 'TOGGLE_CENTER_ALIGN_OPTION';
-export const toggleCenterAlignOption = createAction(TOGGLE_CENTER_ALIGN_OPTION);
+export const toggleCenterAlignOption = () => dispatch => {
+  const state = store.getState();
+  dispatch(setCenterAlignOption(!state.centerAlignGurbani));
+};
```
bogas04:
```suggestion
export const toggleCenterAlignOption = () => (dispatch, getState) => {
  const state = getState();
  dispatch(setCenterAlignOption(!state.centerAlignGurbani));
};
```

bogas04:
Since this is to fix center align related issues, could you also make this change;

Remove Center Align from the search page, as it's useless there.

PR #717, opened 2019-06-20
Self destroying service worker

Gauravjeetsingh:
Adds a self-destroying service worker
Enabled service woker in deploy command
Changed version to 1.1.14


package.json
```diff
@@ -28,11 +28,11 @@
     "deploy:prod": "npm-run-all deploy:prod:*",
     "deploy:prod:build": "NODE_ENV=production npm run build",
     "deploy:prod:clean": "rm -rf .editorconfig .gitignore .github .stylelintrc .vscode stats.json webpack.config.js src LICENSE CONTRIBUTING.md README.md jest",
-    "deploy:prod:pm2": "kill_service_worker=1 pm2 start server --update-env --name=\"sttm-prod\" --node-args=\"-r esm\"",
+    "deploy:prod:pm2": "pm2 start server --update-env --name=\"sttm-prod\" --node-args=\"-r esm\"",
     "deploy:dev": "npm-run-all deploy:dev:*",
     "deploy:dev:build": "npm run deploy:prod:build",
     "deploy:dev:clean": "npm run deploy:prod:clean",
-    "deploy:dev:pm2": "kill_service_worker=1 pm2 start server --update-env --name=\"sttm-dev\" --node-args=\"-r esm\""
+    "deploy:dev:pm2": "pm2 start server --update-env --name=\"sttm-dev\" --node-args=\"-r esm\""
```
bogas04: Instead make it 0 for future use

scripts/generate-sw.js
```diff
@@ -20,4 +30,4 @@ fs.writeFileSync(
       /%CSS_ASSETS/gi,
       `'/assets/css/vendor/foundation.min.css?v=6.2.4', '/assets/css/bundle.css'`
     )
-);
+);*/
```
bogas04: Instead have `sourceFile` variable and let the code be the way it was, only change the value of `sourceFile`

package.json
Gauravjeetsingh: Done.

scripts/generate-sw.js
Gauravjeetsingh: Fixed.

PR #718, opened 2019-06-21
Release 1.1.14

Gauravjeetsingh:
- Shows share links when we select text
- Fixes 404 page error in embed links
- Changes background color of options in light theme
- Fixes scroll position in sundar gutka
- Fixes object object error
- Adds option to center align Gurbani
- **Adds a self-destroying service worker to fix the cache issue.**


bogas04:
Selection thing was also done, add to change log

PR #732, opened 2019-07-22
Adds new fonts prabhki and raaj for Gurbani

ekas:


Gauravjeetsingh:
Currently, if I select any other font, we get this empty space below each pankti. 
Screenshot from 2019-07-22 18-24-56

Increase the spacing only when raaj font is selected.
Also, Instead of changing the margin value, change the line-height of the pankti to avoid overlapping. That will also fix the overlapping with a change in font-size.

ekas:
> Currently, if I select any other font, we get this empty space below each pankti.
> Screenshot from 2019-07-22 18-24-56
> 
> Increase the spacing only when raaj font is selected.
> Also, Instead of changing the margin value, change the line-height of the pankti to avoid overlapping. That will also fix the overlapping with a change in font-size.

@Gauravjeetsingh as you mentioned I have added those changes. Please review.

Gauravjeetsingh:
Hello @ekas

This looks good. Prabhki font is looking good.

There's a slight font issue with raaj, it shows ੴ  twice. Any ideas, how can we handle that?
Screenshot from 2019-07-26 09-23-34

Also, @saintsoldierx veerji, need your thoughts on this as well, I think, we should not give too many variants of Raaj font in options. Should we stick to just one style of raaj font?

ekas:
Screenshot 2019-07-27 at 12 07 36 PM
Screenshot 2019-07-27 at 12 07 53 PM

@Gauravjeetsingh  Could this be the issue.

ekas:
I removed one of the angular brackets.

Gauravjeetsingh:
@ekas Yeah, but we need double angular brackets for other fonts. Can you look up if we can edit characters from font file? If we can set >> to the same character as > in the font, that should work.

ekas:
Okay, let me have a look. @Gauravjeetsingh

PR #754, opened 2019-08-01
Hides controls when we scroll down

Gauravjeetsingh:
chrome-capture


bogas04:
Not quite happy with opacity effect. Maybe we can use something else (Gurmukhi Keyboard/translateY)

src/js/components/Controls.js
```diff
@@ -44,6 +46,24 @@ class Controls extends React.PureComponent {
...
+        }
+      } else if( this.state.lastScrollPos < currentScroll) {
+        this.setState({lastScrollPos: currentScroll});
+
+        if (this.state.showControls &&
+           !this.props.showDisplayOptions &&
+           !this.props.showFontOptions) {
+          this.setState({showControls: false});
+        }
+      }
```
bogas04: Can you read state once and set it once?

```js
const { showDisplayOptions, showFontOptions } = this.props;

this.setState(prevState => {
 const { showControls, lastScrollPos } = prevState;

  if(lastScrollPos > currentScroll) {   
    return {
      lastScrollPos: currentScroll,
      showControls: !showControls
        ? true
        : showControls
    };
  } 

  return {
    lastScrollPos: currentScroll,
    showControls: showControls &&
      !showDisplayOptions &&
      !showFontOptions ?
        false :
        showControls
  };
});
```

Gauravjeetsingh: Thank you for this. Updated the code. Also switched to a flip animation instead of fade.

src/js/components/Controls.js
```diff
@@ -57,7 +82,7 @@ class Controls extends React.PureComponent {
     return (
       <div
         id="controls-wrapper"
-        className={`no-select ${this.state.showBorder ? 'with-border' : ''}`}
+        className={`no-select ${this.state.showBorder ? 'with-border' : ''} ${this.state.showControls ? 'show-controls' : 'hide-controls'}`}
```
bogas04: Use classnames package

src/js/components/Controls.js
```diff
@@ -44,6 +46,29 @@ class Controls extends React.PureComponent {
...
+        }
+        return {
+          lastScrollPos: currentScroll,
+          showControls: showControls &&
+          !showDisplayOptions &&
+          !showFontOptions ?
+          false :
+          showControls
+        };
+      });
```
bogas04: Are you not using prettier? Please set it up

Gauravjeetsingh: Done. :pray:

src/js/components/Controls.js
Gauravjeetsingh: Ah! that clears up the code. Done.

saintsoldierx:
On chrome, sometimes when you scroll up it shows the shadow on the bar, sometimes it doesn't.

Ideally, it should show always. Can this be fixed?

image
 
image

Gauravjeetsingh:
@saintsoldierx veerji! I am unable to reproduce above bug. I scrolled up/down multiple times in chrome. Even on the shabad in your screenshot. Is there a particular scenario when it happens?

saintsoldierx:
Please resolve conflict - then will merge.

PR #757, opened 2019-08-01
Fixes the next ang bug

Gauravjeetsingh:
Clicking on next arrow while reading in ang view goes to a SGGS ang, even when we are on Dasam Granth, this PR fixes that bug.


bogas04:
Search the code for other such cases as well. Nice find.

Gauravjeetsingh:
Found a bunch of em, @bogas04 can you re-check and merge if all looks good?

Issue #752, opened 2019-07-31
Connection timed out error on Ang search


Gauravjeetsingh:
**Describe the bug**
When we select the search type to Ang, and search source to `All Sources`, `Amrit Keertan`, `Bhai Nand laal ji vaaran` or `Rehatnamas & Panthic Sources`, we get Connection timed out.

**To Reproduce**
1. Select search type to Ang
2. Select search source to any of the following:
`All Sources`, `Amrit Keertan`, `Bhai Nand laal ji vaaran` or `Rehatnamas & Panthic Sources`
3. Search for any Ang.

**Expected behavior**
If these sources can't be read by Ang, we should not get these source in the options, when Ang search type is selected.

**Screenshots**
![Screenshot_2019-07-31_12-38-06](https://user-images.githubusercontent.com/1990932/62190917-12961900-b390-11e9-9f0d-ae7c34fce4ad.png)

Issue #767, opened 2019-08-07
Default "ang" search to SGGSJ


saintsoldierx:
**Describe the bug**
The default source for ang should be SGGSJ. Not all sources.

**To Reproduce**
1. Go to 'sikhitothemax.org'
2. Select 'ang'
3. Search for ang 35 for example

**Expected behavior**
Should give a relevant message to say select a source OR default to SGGSJ.

saintsoldierx:
Remove option for "ALL SOURCES" and default to SGGSJ only when ANG is selected. When any other option other than ANG is selected, it should re-show the ALL SOURCES option.

Gauravjeetsingh:
Already created an issue for this #752

tarunsingh5:
@Gauravjeetsingh was this merged to master and deployed?

Gauravjeetsingh:
No, it hasn't been merged to master. There's a PR opened for this release. Need to complete the testing once.

tarunsingh5:
what's our ETA on this - its been pending for almost 2 months

Gauravjeetsingh:
It's ready to be shipped from my side. I and Akal Ustat ji tested it yesterday, and there were no issues found.

Just need someone to merge this PR:
https://github.com/KhalisFoundation/sttm-web/pull/787

Gauravjeetsingh:
Closing it as it's pushed to prod.

PR #769, opened 2019-08-08
Ang search issue

Gauravjeetsingh:
Fixes issue #752 / #767


Gauravjeetsingh:
@bogas04 veerji, using the sources with ang object now from banidb package. Can you please take a look at this, let me know if it requires any other change?

PR #772, opened 2019-08-14
Code refactoring to avoid regression bugs

Gauravjeetsingh:
@bogas04 ji, this is how I am currently thinking about refactoring. Created a new folder api to have all such functions. Let me know if I am on the right track.


src/js/util/api/shabad.js
```diff
@@ -0,0 +1,40 @@
...
+  english: shabad => shabad.raag.english,
+  gurmukhi: shabad => shabad.raag.gurmukhi,
+  unicode: shabad => shabad.raag.unicode,
+};
+
+export const getWriter = {
+  english: shabad => shabad.writer.english,
+  gurmukhi: shabad => shabad.writer.gurmukhi,
+  unicode: shabad => shabad.writer.unicode,
+};
```
bogas04: Getter nomenclature is usually assigned to a function, maybe make these functions?

Gauravjeetsingh: Done.

PR #787, opened 2019-09-12
Release 1.3.0

Gauravjeetsingh:
Changelog:
- Ability to print shabad
- Code optimizations
- Changes hukamnama navigation
- Fixes the source list on search type ang


now.json
```diff
@@ -1,7 +1,8 @@
 {
+  "version": 1,
   "name": "@sttm/web",
   "engines": {
-    "node": "v8.11.2"
+    "node": "v8.11.3"
```
bogas04-swiggy: Let's update to node 12 later sometime. It's gonna LTS in October

src/js/pages/Sync/index.js
```diff
@@ -81,7 +108,7 @@ export default class Sync extends React.PureComponent {
   componentWillUnmount() {
     this._mounted = false;
     if (this.state.connected && this._socket) {
-      this._socket.disconnect();
+      //this._socket.disconnect();
```
bogas04-swiggy: Do we not want this?

Gauravjeetsingh: If we disconnect the socket, we are unable to reconnect to same sync code again, or connect from any other instance of sttm-web with that code. We have to stop and start a new session from the desktop. So, I commented it out for now.

Gauravjeetsingh: Also, the link to sync is not yet exposed. So, we aren't planning to release it with this version. It was merged to dev, because of the ikmanzil demo we have. So that they can use dev link to test it out.

bogas04-swiggy: Here we disconnect only when user leaves the screen. When the screen is reopened, we'll remount.

I understand you might want user to persist the connection, but if we keep it on it might kill the battery on phone and will leave too many open connections on server. Can we instead disconnect it after, say 5 minutes? While persistent connections are cool, they might not scale for our server implementation

Gauravjeetsingh: Yeah, I got that. Even with this I still wasn't able to implement a persistent connection though.

And earlier, if we disconnect the socket from the web, it also disconnected the desktop session. But, I tested it again now, and it doesn't act that way. So uncommenting it now.

Gauravjeetsingh: @bogas04 veerji I am a little confused about this disconnect code. Will continue this conversation on slack. I have commented the code again, after uncommenting it.

tarunsingh5: any update here?

Issue #710, opened 2019-06-13
Unable to open Bhai nand laal ji vaaran by panna


Gauravjeetsingh:
**Describe the bug**
Unable to read Bhai Nand laal ji vaaran by panna.

**To Reproduce**
1. Go to https://www.sikhitothemax.org/shabad?id=30005&q=dAdb&type=1&source=all&highlight=400052
2. Click on the panna link in shabad meta

**Expected behavior**
It should open the panna containing that shabad or give a clear error if we can't read it by panna.

**Screenshots**
![Screenshot from 2019-06-12 18-30-37](https://user-images.githubusercontent.com/1990932/59397085-3221a580-8d40-11e9-8c2c-9455bc707da3.png)

ManjotS:
They don't have pages associated. This is as designed.

Gauravjeetsingh:
Yeah, then we shouldn't show a link to panna in the shabad page, for this source. 

Because current error tells that something isn't working.

saintsoldierx:
If we can find a way to remove that link just for Bhai Nand Laal Ji's vaaraan, that would be great.

PR #803, opened 2019-10-08
Hides the panna link on sources which can't be traversed by Ang

Gauravjeetsingh:
Fixes issue #710


saintsoldierx:
Resolve conflict.

PR #806, opened 2019-10-15
This MR:

gurpreet-hanjra:
1. Disables entire div selection when single word is selected.
2. Fix unique key warning on span in Breadcrumb.

### DIFF
--- a/src/js/components/Baani.js
+++ b/src/js/components/Baani.js
@@ -6,7 +6,7 @@ import Transliteration from './Transliteration';
 import BaaniLine from './BaaniLine';
 import { clickEvent, ACTIONS } from '../util/analytics';
 import { TEXTS, SHABAD_CONTENT_CLASSNAME } from '.././constants';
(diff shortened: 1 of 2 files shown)


gurpreet-hanjra:
> Yeah, that sure fixes it.
> 
> Although, the only thing I had in mind is, that clipboard icon copies the entire pankti, even when we select a single word. @saintsoldierx veerji, any suggestion here?
> 
> Should we keep it like it is now, or we change to copy only the selected words.
> I like the later option, but then there won't be any option to copy entire pankti, without selecting it.

Maybe have another icon along copy to clipboard to copy whole block. Just a suggestion. In this way, user can copy a word or copy whole block if he/she wants.

saintsoldierx:
This is the only icon i can think of for copying just the selection... https://iconscout.com/icon/copy-selection

saintsoldierx:
Let's leave without icon - please resolve conflicts. @Gauravjeetsingh

PR #847, opened 2019-11-16
Add desktop and mobile app links to footer 

saintsoldierx:


Gauravjeetsingh:
@saintsoldierx veerji, it breaks footer on smaller screen.

Screen size somwhere between 1180px and 950px shows footer in two rows.

Screenshot from 2019-11-18 11-26-15

saintsoldierx:
I will fix it in upcoming release. Should I create another @media screen res to fix it? We currently only have two and i have hidden the icons for those res'.

Gauravjeetsingh:
Yeah, another media screen should do it I think.
For this newer media screen, Instead of hiding, what if we change them to smaller icons, without text (Get it on google play..)

Issue #120, opened 2017-03-16
Consider changing to more modern fonts


ManjotS:
ie open sans and the like

@navdeepsinghkhalsa @bogas04 id like your input on this

especially want to do this in #metadata etc

bogas04:
Any decision made on this? @tarunsingh5

ManjotS:
we want to do it its just low priority

saintsoldierx:
@ManjotS what does this entail? Which fonts/text need changing?

ManjotS:
8 think this falls under redesign. I want to use more modern fonts for
English.

saintsoldierx:
We decided to go with Lato (Google fonts)

Gauravjeetsingh:
@saintsoldierx veerji, currently Lato is used only in the footer, and we are using Muli everywhere else. Is this a bug or intentional?

saintsoldierx:
Hanji yes it’s intentional. See discussion in slack #design with Manjot Singh.

PR #857, opened 2019-11-18
Change English font, add hover for search button #120

saintsoldierx:


src/scss/_footer.scss
```diff
@@ -1,4 +1,5 @@
 footer {
+  font-family: 'Lato', Sans Serif;
```
Gauravjeetsingh: This would be font-family: 'Lato', sans-serif;

src/scss/_header.scss
```diff
@@ -131,9 +131,15 @@
...
-
+    transition: 0.3s;
+    &:hover{
+      background-color: $sttm-orange;
+    }
     .dark-mode & {
       background-color: $sttm-blue;
+      &:hover{
+      background-color: $sttm-orange;
+    }
```
Gauravjeetsingh: We are writing the same css code for dark mode again, can we just skip it then. Ideally, it should work on both modes without rewriting.

Issue #815, opened 2019-10-21
Fix a few warnings when deploying


saintsoldierx:
Fix a few warnings when deploying (DEV):-

WARNING in ./src/js/components/BaaniLineActions/index.ts 1:0-69
“export ‘IBaaniLineActionsProps’ was not found in ‘./BaaniLineActions’
@ ./src/js/components/Baani.js
@ ./src/js/components/ShabadContent/ShabadContent.js
@ ./src/js/components/ShabadContent/index.ts
@ ./src/js/pages/Shabad/index.js
@ ./src/js/routes.js
@ ./src/js/Root.js
@ ./src/js/index.js
WARNING in ./src/js/components/Fetch/index.ts 1:0-47
“export ‘IFetchProps’ was not found in ‘./Fetch’
@ ./src/js/pages/PageLoader.js
@ ./src/js/pages/Ang/index.js
@ ./src/js/routes.js
@ ./src/js/Root.js
@ ./src/js/index.js
WARNING in ./src/js/components/RelatedShabads/index.ts 3:0-32
“export ‘IRelatedShabadsProps’ was not found in ‘./RelatedShabads’
@ ./src/js/components/ShabadContent/ShabadContent.js
@ ./src/js/components/ShabadContent/index.ts
@ ./src/js/pages/Shabad/index.js
@ ./src/js/routes.js
@ ./src/js/Root.js
@ ./src/js/index.js
WARNING in ./src/js/features/reducers/index.js 94:9-30
“export ‘TOGGLE_UNICODE_OPTION’ was not found in ‘../actions’
@ ./src/js/features/store/index.js
@ ./src/js/index.js
WARNING in ./src/js/features/reducers/index.js 98:18-39
“export ‘TOGGLE_UNICODE_OPTION’ was not found in ‘../actions’
@ ./src/js/features/store/index.js
@ ./src/js/index.js
WARNING in entrypoint size limit: The following entrypoint(s) combined asset size exceeds the recommended limit (244 KiB). This can impact web performance.
Entrypoints:
 app (280 KiB)
     chunks/vendor-5f4396f0b453524fbb96.js
     app-5f4396f0b453524fbb96.js

rajkambo:
@saintsoldierx mind if I take this one on?

saintsoldierx:
@rajkambo Please do! :)

PR #877, opened 2019-12-03
Fix local deploy warnings

rajkambo:

### DIFF
--- a/src/js/components/BaaniLineActions/index.ts
+++ b/src/js/components/BaaniLineActions/index.ts
@@ -1,1 +1,4 @@
-export { default, IBaaniLineActionsProps } from './BaaniLineActions';
+import BaaniLineActions, { IBaaniLineActionsProps } from './BaaniLineActions';
+
+export type IBaaniLineActionsProps = IBaaniLineActionsProps;
+export default BaaniLineActions;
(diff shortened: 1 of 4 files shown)


rajkambo:
Took a while to find, but https://github.com/webpack/webpack/issues/7378 had a workaround (using `export type`) for how to export interfaces properly. Going to run the sanity tests after lunch so please don't merge yet.

rajkambo:
@saintsoldierx I wasn't able to reproduce the last warning in the ticket https://github.com/KhalisFoundation/sttm-web/issues/815:

```
WARNING in entrypoint size limit: The following entrypoint(s) combined asset size exceeds the recommended limit (244 KiB). This can impact web performance.
Entrypoints:
app (280 KiB)
chunks/vendor-5f4396f0b453524fbb96.js
app-5f4396f0b453524fbb96.js
```

and I checked the logs of the above dev deploy and didn't find it there either? so not sure if that is still an issue?

saintsoldierx:
Thanks @rajkambo ! No worries about the last issue. Just making sure - this won't cause issues with unicode right?

rajkambo:
No problem! And sorry forgot to comment about that. But it looks like we don’t have the “toggle Unicode” button anymore under fonts rights? We only have the option to set to Unicode or the other fonts, so that kind of rendered the toggle case unnecessary right?

saintsoldierx:
Ahh yes you're right - it's not a toggle button, it is a list item under the FONTS dropdown. Thanks again @rajkambo !

PR #886, opened 2019-12-12
CSS Design Updates: Hamburger menu, index page

saintsoldierx:


Gauravjeetsingh:
Yay! finally a design update. I like the translucent effect in menu. :+1: 

Just a few thoughts about this menu design:

- This feels a little congested on mobile screen, without the upper margin.
Screenshot from 2019-12-12 11-46-12

- In dark mode, the color of the menu icon and the back icon isn't visible, what if we change them to orangish-yellow(#f39c1d) color? I think it's always been less visible, now that we are redesigning, we can explore this.
compar3

- In light mode, it's a little difficult to read because of less contrast between the background color and text. What if we keep it white or a color similar to it, like rgb(240, 248, 255)?
compare4

src/scss/_header.scss
```diff
@@ -392,7 +392,7 @@
 @media screen and (max-width: 55.99em) {
   .top-bar-title {
     float: none;
-    margin: 50px 0 0;
+    margin: 0 0;
```
Gauravjeetsingh: We can just write margin: 0 here.

PR #892, opened 2019-12-15
Updates to styles and static pages

saintsoldierx:
* [X ]  Passed [sanity tests](http://bit.ly/sttm-sanity-tests).


src/js/pages/About/index.js
```diff
@@ -9,35 +9,39 @@ export default class About extends React.PureComponent {
...
+          handheld devices.
           <br />
           <br />
           SikhiToTheMax is now developed by Khalis Foundation as seva to the
-          panth and is working hard to ensure it lives up to its name.
+          Panth and are working hard to ensure it remains the most accurate Gurbani application.
           <br />
           <br />
-          <span style={{ fontWeight: 500 }}>Website:</span>
+          <span style={{ fontWeight: 500 }}>Website: </span>
```
Gauravjeetsingh: Should we keep all the styles in the scss files for consistency?

src/scss/_shabad-page.scss
```diff
@@ -15,11 +15,11 @@ blockquote {
 
 #metadata {
   align-items: center;
-  border: 1px solid $sttm-light-grey;
+  background-color: #e9e9e9;
```
Gauravjeetsingh: If the current colors we have as variables, doesn't work here, maybe we can add some new colors to it and use a variable here instead of hex code.

List of current colors is at `sttm-web/src/scss/_colours.scss`

src/scss/_spinner.scss
```diff
@@ -1,40 +1,31 @@
...
+    box-shadow: 0 0 0 35px rgba(0, 0, 0, 0);
   }
 }
 
 @media screen and (max-width: 45.999999em) {
   .spinner {
-    background-size: 100px 100px;
-    height: 100px;
-    width: 100px;
+    background-size: 150px 150px;
```
Gauravjeetsingh: Can we use `background-size: 150px` here?

PR #950, opened 2020-01-23
Fixes issues with sttm-banner

Gauravjeetsingh:
Changed the notification toggle logic. Tested on safari and chrome browser.

### DIFF
--- a/src/js/components/Banner.js
+++ b/src/js/components/Banner.js
@@ -28,6 +28,7 @@ export default class Banner extends React.PureComponent {
         const row = messages.rows[0];
         this.setState({ index: $index });
         const hasSeenMsg = getBooleanFromLocalStorage(`SeenBanner-${$date}-${$index}`, false);
+        console.log(hasSeenMsg);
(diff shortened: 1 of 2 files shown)


src/js/components/Banner.js
```diff
@@ -1,88 +1,62 @@
...
     date: 0,
-    index: 0,
   };
 
   componentDidMount() {
-    const d = new Date();
-    const $mysqlDate = d.getFullYear() + '-' + (d.getMonth() + 1) + '-' + d.getDate() + ' ' + d.getHours() + ':' + d.getMinutes() + ":" + d.getSeconds();
-    const $date = $mysqlDate.slice(0, 10);
+    const currentDate = new Date();
+    const date = currentDate.toISOString().slice(0, 10);
```
AkalUstat: `const date = currentDate.toISOString().slice(0, 10);`
The toISOString function will auto go to the internation dateline time and will therefore not show messages based on user's local time. Might be a problem

Gauravjeetsingh: Ah! got it. Will update. Thanks :)

Issue #885, opened 2019-12-11
Short Ang URL always opens the ang in Guru Granth Sahib ji


Gauravjeetsingh:
**Describe the bug**
Short URL of Ang 1 of Guru Granth Sahib Ji: http://sttm.co/a/1
Short URL of Ang 1 of Dasam Granth Sahib: http://sttm.co/a/1
Short URL of Ang 1 of Bhai Gurdas Ji Vaaran: http://sttm.co/a/1

All the short URLs of ang view are the same, and thus they always redirect to first ang of Guru Granth Sahib ji.

**To Reproduce**
1. Go to https://www.sikhitothemax.org/ang?ang=1&source=B
2. Click on copy icon next to short URL
3. Open the copied URL in the new tab
4. It will open the first ang of Guru Granth Sahib ji instead of Bhai Gurdas Ji Vaaran

**Expected behavior**
Short url should have a source parameter, to open the Ang in specified source.

saintsoldierx:
Short URL of Ang 1 of Guru Granth Sahib Ji: http://sttm.co/a/1
Short URL of Ang 1 of Dasam Granth Sahib: http://sttm.co/d/1
Short URL of Ang 1 of Bhai Gurdas Ji Vaaran: http://sttm.co/b/1
Short URL of Ang 1 of Bhai Gurdas Singh Ji Vaaran: http://sttm.co/s/1

ekas:
@saintsoldierx Veerji, Just wanted to check with you whether there is a portal to configure the shorten urls against the DNS names. Because, I was trying to locate the shorten urls controller layer which I wasn't able to find.

ManjotS:
https://github.com/KhalisFoundation/sttm-shortener

saintsoldierx:
UPDATED:

Short URL of Ang 1 of Guru Granth Sahib Ji: http://sttm.co/**g**/1
Short URL of Ang 1 of Dasam Granth Sahib: http://sttm.co/**d**/1
Short URL of Ang 1 of Bhai Gurdas Ji Vaaran: http://sttm.co/**v**/1
Short URL of Ang 1 of Bhai Gurdas Singh Ji Vaaran: http://sttm.co/**gs**/1

Short URL of Sunder Gutka is now http://sttm.co/**sg**/

Also, update source value on STTM of Bhai Gurdas Jee Vaaran to 'V' and
Bhai Gurdas Singh Vaaran to 'GS' so 
[...]

PR #1007, opened 2020-04-07
Fixes issue #885

Gauravjeetsingh:
Updates the short url of ang views.

### DIFF
--- a/src/js/util/index.js
+++ b/src/js/util/index.js
@@ -70,6 +70,17 @@ export function getQueryParams(str = document.location.search) {
  *
  * @param {string} url
  */
+
+const shortURLSource = (source) => {
+  const urlSource = {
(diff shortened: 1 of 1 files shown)


saintsoldierx:
Can you clarify what the change is @Gauravjeetsingh ? Because in the short url, i still see Dasam Bani and Bhai Gurdas Jee vaars using letter S. Also, Sundar Gutka reads sundar-gutka, where is should be SG...

image

image

Gauravjeetsingh:
@saintsoldierx veerji, you are currently looking at short URLs of Shabads. All Shabads would have a short URL with s.
This issue was only for Ang view. Go to ang view -> Change source, you will get different short URL. Earlier it was the same for all sources.

PR #1044, opened 2020-04-27
[feat]: New Shortcuts

AkalUstat:
https://drive.google.com/open?id=1ILu1IqxxnFCV8Rs4BAHHC0qTl3CvTdx_6p-EeA--lKY


Gauravjeetsingh:
@AkalUstat ji, I am unable to use the keyboard shortcuts. 

Testing it here
https://khalis-herok-shortcuts-jwps3u1.herokuapp.com/shabad?id=903

Do I need to enable it from somewhere or something?

AkalUstat:
You have to be clicked on the shabad viewer

Gauravjeetsingh:
Can you check once, it's not working for me in hukamnama page as well. Also, I see the blue outline everywhere.
image

AkalUstat:
@saintsoldierx @Gauravjeetsingh changed translation key codes to this:

english: `shit+e'
teeka: `shift+t`
spanish: `shift+s`

as the alt+e was causing issues

AkalUstat:
@saintsoldierx I will create some user info for this stuff in a separate PR

AkalUstat:
> Looks good to me. Awesome work @AkalUstat
> 
> I only have a concern about the keyboard shortcuts we use
> For Shahmukhi translation toggle, we use: ctrl + alt + shift + t + s, that's five keys. It takes both hands to toggle that. It can be done with one hand as well, but all fingers are used and are put in a weird position.
> 
> Also, @AkalUstat are we using the same shortcut keys that are currently in desktop, if not, we should.
> 
> @saintsoldierx @maneetpaul @ManjotS veerjis, can any of you you review the shortcut keys used here if you haven't yet.

1. Yea I agree with the translit toggle; problem is if we do cmd+alt+e (for english translit,  without the shift and the t) due to an issue with alt +_ e it messes up all the other shortcuts. I can try to debug and see if i can come up with resolution

2. Only desktop one that matches is Cmd + /. Desktop does not have shortcuts to modify display items. Waiting on react refactor to add these to desktop

Gauravjeetsingh:
Found a small bug
For toggling Shahmukhi transliteration, one has to take the hand up from all five keys and press again. 
For example with Hindi transliteration, I just pick my finger from last key and press again and it works.

AkalUstat:
Interesting...I will try to debug as well

AkalUstat:
:ti

AkalUstat:
There is error when the alt key is used for shortcuts in the library, documented here https://github.com/greena13/react-hotkeys/issues/269 so we need new mappings for translit characters to avoid the long shortcuts

src/js/Shortcuts.js
```diff
@@ -0,0 +1,98 @@
...
+    store.dispatch({type: 'SET_TRANSLITERATION_LANGUAGES', payload: toggleItemInArray('shahmukhi', state.transliterationLanguages)})
+  },
+  toggleHinTranslit: () => {
+    const state = store.getState();
+    store.dispatch({type: 'SET_TRANSLITERATION_LANGUAGES', payload: toggleItemInArray('hindi', state.transliterationLanguages)})
+  }
+}
+
+// export default Shortcuts;
+export {ViewerShortcutHanders, ViewerShortcuts, GlobalHandlers, GlobalShortcuts}
```
plug-n-play, line 98: Great. We should also document the keyboard shortcuts both in the ReadMe.md and somewhere in the application - may be in the footer? @saintsoldierx thoughts?

AkalUstat: I was thinking the help menu (but for another pr) and then a quick help toggle at the top bar

AkalUstat:
New transliteration keymaps: 
`alt+shift+e`: english
`alt+shift+s`: shahmukhi
`alt+shift+h` hindi

AkalUstat:
added shift + l shortcut for larivar assist

Issue #1057, opened 2020-05-12
History Support for Bani Controller


maneetpaul:
Bani Controller is no longer displaying history like it used to when we launched. 

This history should stay in Sync with the history on desktop and vice versa

saintsoldierx:
I still don't think history has been added to Bani Controller? @Gauravjeetsingh @maneetpaul ?

gurjit03:
@Gauravjeetsingh please can you update on this one or close if done.

gurjit03:
@Gauravjeetsingh veerji, any update on this ? Should we keep it open or close it?

Gauravjeetsingh:
Yeah, need some clarity on this one. Is this about adding history of shabads or history of which verses are read in a shabad?

If its about verses, that should be working, and we can work out few edge cases, where it fails. For history of shabads, I dont remember we ever had that.

gurjit03:
@Gauravjeetsingh sure, can we try to get this needed information, so we can proceed further accordingly.

saintsoldierx:
@gurjit03 I would say both use cases are valid

Issue #1054, opened 2020-05-12
Sundar Gutka Support for Bani Controller


maneetpaul:
If desktop opens a Sundar Gutka Bani, it should be able to be controlled on Bani Controller. 

Additionally, we will eventually be adding Sundar Gutka icon to Bani Controller itself, so let's make sure we future-proof this so this can be an easy add. Meaning, if a Sundar Gutka Bani is selected on web, it should be able to synced back to desktop.

PR #1060, opened 2020-05-18
Two-way sync with Bani & History support

Gauravjeetsingh:


Gauravjeetsingh:
Fixes issue #1057 #1054

PR #1064, opened 2020-05-21
Input validations

gurjit03:

image

### DIFF
--- a/src/js/pages/Home/index.js
+++ b/src/js/pages/Home/index.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { SOURCES, SEARCH_TYPES, TYPES, SOURCES_WITH_ANG, DOODLE } from '../../constants';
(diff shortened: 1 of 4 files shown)


src/js/pages/Home/index.js
```diff
@@ -92,6 +92,8 @@ export default class Home extends React.PureComponent {
                         placeholder={placeholder}
                         title={title}
                         pattern={pattern}
+                        min={name === 'ang' && 1}
+                        max={name === 'ang' && 1430}
```
Gauravjeetsingh, line 96: We can't set the max value to 1430 for each source. Not all sources have 1430 as the total number of angs. For instance, Dasam Granth only has 1428 angs. So this still gives an error
https://khalis-herok-input-vali-tfzhs5.herokuapp.com/ang?ang=1430&source=D

I would suggest creating an object in constants.ts with total angs of each source, and use that here, instead of writing it directly in code.

gurjit03: yeah sounds good. can you share details of where I can get the max angs of each source ?

Gauravjeetsingh:
LGTM :+1:

PR #1065, opened 2020-05-21
adds freshwork widget

gurjit03:

image

**Removes in the fullscreen view**
image


Gauravjeetsingh:
@saintsoldierx veerji, if we add the support button we don't need the feedback button then in the footer, right?

saintsoldierx:
Leave feedback for time being - STTM desktop feedback is still being routed through it i believe.

PR #1067, opened 2020-05-22
Improve related shabads

gurjit03:


Gauravjeetsingh:
This is looking good @gurjit03 :100: It needs some polishing, which I believe you already know.

Can you fix the following things as well, I think these are issues in previous design as well:
- The text of Gurbani in related Shabads gets bolder on the small screen, let's keep the font-weight same, keep it normal as it is on a bigger screen.
- I don't think we need to show translation and transliteration of the related Shabads, show only gurbani.

saintsoldierx:
@gurjit03 It looks sooo good so far. Yes, please do changes @Gauravjeetsingh mentioned and will approve.

saintsoldierx:
@gurjit03 i would maybe leave the translation showing - but remove transliteration. Translation makes it interesting to see what words matched etc.

saintsoldierx:
When they hover over the right bar, is there a way to show ALT TEXT which will say something like "45% match" or whatever?

saintsoldierx:
image

- Results need to be ordered so highest matches at top (see screenshot).
- Percentage needs to be whole numbers when hovering over ie. 45% not 45.75%
- More text should read "Show More Results" in a button form in the center of the page.

saintsoldierx:
image

Almost there! So i think functionality and sorting is good. Issue I see is that the bars on left side showing the percentage look like they are off (eg. in screenshot, both in red box are 45% matching but one bar looks fuller than the other). This is because the height of the boxes are different. Is it possible we can make the height of the percentage bar as static so it's always a certain height, whatever the minimum height of the box is?

saintsoldierx:
.relatedShabadAvgRating 
height: 150px

Try that ^

saintsoldierx:
I would like the button to be like a proper button. CSS:-

image

image

HOVER BG COLOR: #00a6ff (include a transition so it fades in/out on/off hover)

saintsoldierx:
Bar should be at the top , not bottom

https://files.slack.com/files-pri/T03QFACU3-F014HCGUR0W/image.png

PR #1072, opened 2020-05-28
Uses on heroku env variable

Gauravjeetsingh:
Added another env variable ON_HEROKU and switches to PORT env variable for heroku.


Gauravjeetsingh:
I added a condition to check if the deployment is from Heroku.
If it's from Heroku, it'll use the port from env variable, else it will use 8081 for development and 8080 for production. I have tested it on heroku, let me know how to test it on our servers.

PR #1079, opened 2020-06-03
Feature auto scroll 

gurjit03:

 [ a - is shortcut ]
In paused state
image

In scrolling state
image

image


saintsoldierx:
image

saintsoldierx:
image
For mobile devices

saintsoldierx:
@gurjit03 please see designs above for auto scroll.

src/scss/_display-controls.scss
```diff
@@ -12,9 +12,9 @@
   transition: transform 0.2s;
   z-index: 1;
 
-  .dark-mode & {
-    background-color: $sttm-light-black;
-  }
+  // .dark-mode & {
```
Gauravjeetsingh, line 15: We need this here, currently, dark mode is like this

image

gurjit03: ahh good catch thanks !

saintsoldierx:
It doesn't seem to work when i press play to scroll. Moves page a little then stops.

Also, I think it may look nice if we change opacity of the entire panel to 0.5 (.floatingActions) and on hover, change it to 1.0 ?

gurjit03:
@saintsoldierx yeah was thinking the same (for the opacity of the floating icons), need to check the scrolling behaviour was working for me and @Gauravjeetsingh.

Gauravjeetsingh:
@gurjit03 I think @saintsoldierx veerji meant, that auto-scroll stops if we manually scroll while the auto-scroll is active.

This, I can reproduce that from my Linux machine as well.

PR #1082, opened 2020-06-03
Translation transliteration fontsize

gurjit03:

Sizes for translation and transliteration are being controlled now with fontsize

image


saintsoldierx:
@gurjit03 i would like separate font-size controller for Gurmukhi and English translations please.

PR #1086, opened 2020-06-05
makes the display list opens/close on clicking anywhere on document

gurjit03:

This PR fixes the issue with multiselect not being closing on clicking anywhere on the document.

### DIFF
--- a/src/js/components/MultiSelect.js
+++ b/src/js/components/MultiSelect.js
@@ -1,46 +1,86 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { DownArrowIcon } from '@/components/Icons/CustomIcons';
-
-export class MultiSelect extends React.PureComponent {
+export class MultiSelect extends React.Component {
   static propTypes = {
     collections: PropTypes.array.isRequired,
     dropdownLabel: PropTypes.string.isRequired,
   };
+  constructor() {
+    super();
+    this.state = {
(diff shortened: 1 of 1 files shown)


src/js/components/MultiSelect.js
```diff
@@ -1,14 +1,44 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { DownArrowIcon } from '@/components/Icons/CustomIcons';
-
 export class MultiSelect extends React.PureComponent {
   static propTypes = {
     collections: PropTypes.array.isRequired,
     dropdownLabel: PropTypes.string.isRequired,
   };
+  constructor() {
```
plug-n-play, line 9: Converting this to a functional component is better with Hooks. Wherever possible we should try to use Functional components with Hooks for new implementations

plug-n-play, line 9: Converting this to a functional component is better with Hooks. Wherever possible we should try to use Functional components with Hooks for new implementations

gurjit03: @plug-n-play tried using hooks, but seems we are on older version of react

plug-n-play: Go ahead and update to latest. There are no breaking changes

gurjit03: :+1:

PR #1088, opened 2020-06-05
adds border-radius to progress bar

gurjit03:

A very small changes I missed it last time.
image


Gauravjeetsingh:
It's good, just one very minor change.

Instead of giving a border-radius on all four sides, can you change it to border-top-left-radius and border-top-right-radius?

That way it will only have a border on the top, and not at the bottom.

Also, we have border radius of 2 px everywhere else now. Let's use the same value here as well.

PR #1095, opened 2020-06-07
Shortcut fixes

AkalUstat:

Fixes for shortcut:

- Search shortcut now clears the search bar
- using center align shortcut no longer opens the console


AkalUstat:
is it possible that instead of repeating code we just have the action handlers take care of this?

AkalUstat:
Ok so i have done added that functionality to shortcut and allocated the auto toggling nature to the reducer (and so have removed the manual toggling from the controls)

package.json
```diff
@@ -126,4 +126,4 @@
       "pre-push": "npm run pre-push-message && npm run test"
     }
   }
-}
+}
```
plug-n-play, line 129: image
Please fix this

AkalUstat: what does this mean?

saintsoldierx: @AkalUstat at the end of the file, add a new empty line.

Issue #1097, opened 2020-06-07
Unicode searching doesn't highlight correctly.


saintsoldierx:
![image](https://user-images.githubusercontent.com/27791330/83982154-f381e500-a8d8-11ea-86c3-b0186b5e935b.png)

AkalUstat:
I can do this 🤚

Gauravjeetsingh:
Unable to reproduce on dev, tried changing the font to Unicode as well.
Anything else I should be doing to reproduce it @saintsoldierx veerji?

![Screenshot from 2020-06-08 07-41-27](https://user-images.githubusercontent.com/1990932/83986733-79df0b00-a95b-11ea-8395-5fe6d644c523.png)

AkalUstat:
looks to be an issue on ios devices. Am experiencing this on iPhone Xr as well.

AkalUstat:
more specifically safari-based browsers. This issue also appears on mac safari

PR #1101, opened 2020-06-08
Fix: Unicode highlighting being one letter off

AkalUstat:
## The Problem
for some reason, with a unicode query, the last highlight index came out as -1, meaning that the last letter did not get highlighted. Verified through React Devtools

## The Fix
to keep the results consistent, I passed the query through the anvaad unicode function with reverse set to true, so that unicode is converted to ascii. Normal queries will pass through unaffected. This way, the results are consistent regardless of english or unicode.

### DIFF
--- a/src/js/util/index.js
+++ b/src/js/util/index.js
@@ -1,4 +1,5 @@
 /* globals API_URL */
+import { unicode } from 'anvaad-js'
 import {
   DEFAULT_SEARCH_TYPE,
   DEFAULT_SEARCH_SOURCE,
@@ -317,7 +318,7 @@ export const getHighlightIndices = (baani, query, type) => {
   let mainQuery = query.replace(/"/g, '');
 
   //Handles - search operator
-  mainQuery = mainQuery.replace(/[-][ ,\w,),(]*/g, '');
+  mainQuery = unicode(mainQuery.replace(/[-][ ,\w,),(]*/g, ''), true);
 
   if (baani === null) {
     return [start, end];
--- a/package-lock.json
+++ b/package-lock.json
@@ -1706,6 +1706,15 @@
         "color-convert": "^1.9.0"
       }
     },
+    "anvaad-js": {
+      "version": "1.4.2",
+      "resolved": "https://registry.npmjs.org/anvaad-js/-/anvaad-js-1.4.2.tgz",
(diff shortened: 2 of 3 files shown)


AkalUstat:
there are a ton of changes but that is due to the package-lock.json when i installed anvaad-js. Otherwise its a 1 line change

src/js/util/index.js
```diff
@@ -317,7 +318,7 @@ export const getHighlightIndices = (baani, query, type) => {
   let mainQuery = query.replace(/"/g, '');
 
   //Handles - search operator
-  mainQuery = mainQuery.replace(/[-][ ,\w,),(]*/g, '');
+  mainQuery = unicode(mainQuery.replace(/[-][ ,\w,),(]*/g, ''), true);
```
Gauravjeetsingh, line 321: @AkalUstat Why are we converting the search query to unicode here? 

We never used anvaad js directly, API was supposed to use that for us until now.

Gauravjeetsingh: Got the answer from your above description. Just thinking if this is the best way to solve it.

What if we don't convert the search query to unicode ever? When switching to unicode, everything else switches, but not the search query.

AkalUstat: also i really didnt bother to see what was happening with the highlighted index generator i just plugged this in cause it was the first solution i thought of

Gauravjeetsingh: Okay, let me take a look at the highlighted index generator, possibly an error with the regex we use for that. If it's solvable from there, I'd prefer that.

saintsoldierx: any update @Gauravjeetsingh ?

Gauravjeetsingh: I am not able to reproduce this issue on my laptop, and safari on iPhone X

But the query in autocomplete is never Unicode. So this PR fix doesn't affect that. And as per the screenshot in issue #1097, the highlight is off by one word in autocomplete as well. 

I think it's safe to close this PR **without merging**, but the ticket can stay open. I need to investigate more.

AkalUstat: - this is if the user manually enters unicode

tarunsingh5: is this still an issue?

PR #1114, opened 2020-06-12
Feature paragraph mode

gurjit03:

Split view paragraph mode
image

Non split view paragraph mode
image

New Icon for paragraph mode
image


src/js/components/Baani.js
```diff
@@ -109,6 +124,7 @@ export default class Baani extends React.PureComponent {
 
   showShare = e => {
     const currentShare = document.querySelector('.showShare');
+    console.log(currentShare, " current share..");
```
plug-n-play, line 127: let's remove this

saintsoldierx: @gurjit03 agree with Amandeep Singh

saintsoldierx:
If i go to SG and then click paragraph mode, then search for shabad, it still keeps paragraph mode. see example

image

src/js/components/Baani.js
gurjit03: :+1: def

saintsoldierx:
Make sure, if it autoscroll is on and scrolling, if i scroll manually up or down the page, it should not pause. It should keep going.
Seems like it's working on firefox but speed setting doesn't work or is very very slow.

saintsoldierx:
Hukamnama, random shabad, and pages from Index also still show paragraph. They should be disabled.

PR #1157, opened 2020-07-07
Fix larivaar assist color

gurjit03:

This PR fixes larivaar assist color on darkmode and z-index issue with controls list


Gauravjeetsingh:
@gurjit03 veerji, I think we need the color to shift to white on dark mode, instead of black on assist strength 0.

image

gurjit03:
@Gauravjeetsingh yeah right about that

PR #1167, opened 2020-07-13
Add descriptions to SGGS raags and some changes to the index page look

yudhv:

### DIFF
--- a/src/js/pages/GranthIndex/components/GranthIndex.tsx
+++ b/src/js/pages/GranthIndex/components/GranthIndex.tsx
@@ -58,7 +58,7 @@ export const GranthIndex: React.FC<IGranthIndexProps> = ({ location }) => {
         (
           {
             name,
-            pages: [from, to],
+            ang: [from, to],
             source: localSource,
             description,
             subSets,
--- a/src/js/pages/GranthIndex/components/GranthIndex.tsx
+++ b/src/js/pages/GranthIndex/components/GranthIndex.tsx
@@ -150,15 +150,11 @@ export const GranthIndex: React.FC<IGranthIndexProps> = ({ location }) => {
             <div id="help">
               <div id="sidebar">
                 <ul>
-                  {Object.entries(raagIndices).map(
-                    ([key, { name: granthName }]) => (
-                      <li key={key}>
-                        <details>
-                          <summary>{granthName}</summary>
-                        </details>
-                      </li>
-                    )
-                  )}
+                  {Object.entries(raagIndices).map(([key, { name }]) => (
+                    <li key={key}>
+                      <a href={`#${sanitizeHash(name)}`}>{name}</a>
+                    </li>
+                  ))}
                   {/* Amrit keertan Index */}
                   <li>
                     <a href="#amritKeertan">Amrit Keertan</a>
--- a/src/js/pages/GranthIndex/components/GranthIndex.tsx
+++ b/src/js/pages/GranthIndex/components/GranthIndex.tsx
@@ -1,38 +1,147 @@
-import React, { useEffect, MouseEvent } from 'react';
+import React, { useEffect, useState, useCallback } from 'react';
 import { Link } from 'react-router-dom';
 import { Location } from 'history';
 
 import BreadCrumb from '@/components/Breadcrumb';
 import Fetch from '@/components/Fetch';
-import { raagIndices } from '../constants/raag-indices';
+import { raagIndices, IRaagIndex } from '../constants/raag-indices';
 import { toAngURL } from '../../../util';
 import { pageView } from '../../../util/analytics';
 import { AmritKeertanIndex } from './AmritKeertanIndex';
-import { sanitizeHash, scrollToHash } from '../util';
+import { sanitizeHash, scrollToHash, desanitzeHash } from '../util';
 import { TEXTS } from '../../../constants';
 
 interface IGranthIndexProps {
-  location: Location
+  location: Location;
 }
+
+const GRANTH_NAMES = raagIndices.map(({ name }) => name);
 
 export const GranthIndex: React.FC<IGranthIndexProps> = ({ location }) => {
   const { hash } = location;
+
+  const [selection, setSelection] = useState<string[]>(() =>
+    hash ? desanitzeHash(hash) : [GRANTH_NAMES[0]]
+  );
+
+  // Helps check if a list item is selected to show subSets and description
+  const isSelected = useCallback(
+    (name: string) => {
+      if (selection.includes(name)) {
+        return true;
+      }
+    },
+    [selection]
+  );
+
+  // Called onClick to show description and subSets
+  const changeSelection = useCallback(
+    (path: string[]) => {
+      if (path.join() == selection.join()) {
+        path.pop();
+      }
+      setSelection(path);
+    },
+    [selection]
+  );
 
   useEffect(() => {
     pageView('/index');
     // making sure the element is rendered on DOM before we get to that
     setTimeout(() => scrollToHash(hash), 800);
-  }, [])
+  }, []);
+
+  // Recursive function that keeps churning lists until no subSets are left
+  // path is a list that helps keep the parents open when a child is clicked upon
+  const renderList = useCallback(
+    (
+      subSets: IRaagIndex[],
+      source: IRaagIndex['source'] = 'G',
+      path: string[] = []
+    ) => {
+      return subSets.map(
+        (
+          {
+            name,
+            pages: [from, to],
+            source: localSource,
+            description,
+            subSets,
+            highlight,
+          },
+          i
+        ) => {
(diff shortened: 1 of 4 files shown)


src/js/pages/GranthIndex/util/sanitize-hash.ts
```diff
@@ -1 +1,5 @@
-export const sanitizeHash = (...args) => args.map(a => a.replace(/ /gi, '')).join('-');
+export const sanitizeHash = (...args) =>
+  args.map((a) => a.replace(/ /gi, '-')).join('_');
+
+export const desanitzeHash = (hash: string) =>
```
yudhv, line 4: Added this to get back the path of the last clicked list item. With the path available, we should be able to programatically open the list item and scroll to it

yudhv:
Because the descriptions are content-heavy, I had to make the accordion-like functionality to keep the length of the page short. Please suggest if a better approach is warranted.

src/js/pages/GranthIndex/components/GranthIndex.tsx
```diff
@@ -1,38 +1,150 @@
...
   const { hash } = location;
 
+  const [selection, setSelection] = useState<string[]>(() =>
+    hash ? desanitzeHash(hash) : [GRANTH_NAMES[0]]
+  );
+
+  // Helps check if a list item is selected to show subSets and description
+  const isSelected = useCallback(
+    (name: string) => {
+      if (selection.includes(name)) {
```
gurjit03, line 30: @yudhv prolly we can directly return `selection.includes(name)` it gonna return boolean.

src/js/pages/GranthIndex/components/GranthIndex.tsx
```diff
@@ -1,38 +1,150 @@
...
+            // Give a default value if highlight is undefined
+            highlight = 0,
+          },
+          i
+        ) => {
+          if (localSource == null) {
+            localSource = source;
+          } else {
+            source = localSource;
+          }
```
gurjit03, line 79: @yudhv not sure I am following here properly. Couldn't see the source variable being used.

src/js/pages/GranthIndex/components/GranthIndex.tsx
```diff
@@ -1,38 +1,150 @@
...
+                  </div>
+                  {isSelected(name) && (
+                    <div className="index-list--description">{description}</div>
+                  )}
+                </div>
+              </li>
+              {isSelected(name) && subSets && subSets.length > 0
+                ? renderList(subSets, localSource, path.concat([name]))
+                : null}
+            </div>
```
gurjit03, line 131: @yudhv veerji, here we are opening the sublist from the main list item [if we have clicked on it], right?

src/js/pages/GranthIndex/components/GranthIndex.tsx
```diff
@@ -1,38 +1,150 @@
...
+    [selection]
+  );
+
+  // Called onClick to show description and subSets
+  const changeSelection = useCallback(
+    (path: string[]) => {
+      if (path.join() == selection.join()) {
+        path.pop();
+      }
+      setSelection(path);
```
gurjit03, line 43: @yudhv veerji, as much as I understood, we doing this pop() here to toggle the state of that list item. If it's open, and we click on that, we are moving it using pop? if so, that would work on last level list item as well ?

src/js/pages/GranthIndex/components/GranthIndex.tsx
yudhv: yes

src/js/pages/GranthIndex/components/GranthIndex.tsx
yudhv: @gurjit03 that is correct. We only pop when the current selection and onClicked selection are the same, to collapse the clicked list-item. Just checked, it works if a description is added to the vaars (i.e. it opens and closes as expected when clicked) which are the last item in the heirarchy.

src/js/pages/GranthIndex/components/GranthIndex.tsx
yudhv: Thank you. I removed this chunk and tested the source functionality.

AkalUstat:
Just a minor nitpick could not use words like holy book and page?

yudhv:
> Just a minor nitpick could not use words like holy book and page?

@AkalUstat I kept it that way for easier understanding from a new comer's perspective. Granth and Ang aren't English words. If you still want them changed, please let me know what you would prefer.

saintsoldierx:
I really love this idea, definitely a big improvement to the index page. We do need to take text out of the files for the descriptions etc and put it in the api. I have already communicated this to Manjot Singh too who will work on this on the api level. We also want to do this for localization as thats also a project we are working on. Also, we do need to be careful about some of the entries on index. For example, Rehras etc because the compilations of Rehras do vary for different sampardas and also, Banis like Chaupai Sahib wouldn't be in SGGSJ index so it's not really correct. But this is a great starting point.

saintsoldierx:
> > Just a minor nitpick could not use words like holy book and page?
> 
> @AkalUstat I kept it that way for easier understanding from a new comer's perspective. Granth and Ang aren't English words. If you still want them changed, please let me know what you would prefer.

For Granth, it's ok to use "sacred scriptures" and for ang, i would put ang and maybe add a tooltip to explain that "ang directly means limb and can be seen as pages of Sri Guru Granth Sahib Ji" or something along those lines.

yudhv:
> I really love this idea, definitely a big improvement to the index page. We do need to take text out of the files for the descriptions etc and put it in the api. I have already communicated this to Manjot Singh too who will work on this on the api level. We also want to do this for localization as thats also a project we are working on. Also, we do need to be careful about some of the entries on index. For example, Rehras etc because the compilations of Rehras do vary for different sampardas and also, Banis like Chaupai Sahib wouldn't be in SGGSJ index so it's not really correct. But this is a great starting point.

Thank you for checking it out @saintsoldierx . I couldn't find a reference to Chaupai Sahib in the SGGS index. I only saw it in the Dasam Granth index. Regarding Rehras Sahib, it seems we should refer to it as `So Dar` since that is the starting point up until the point that Sohila Sahib starts: https://www.sikhiwiki.org/index.php/Index_of_Guru_Granth_Sahib

So, I'll make the following changes if you all give the go ahead: 
* Remove references to `holy book` and `page`
* Change Rehas to So Dar
* Explore the tooltip option and see if there's a fast to way to implement

@saintsoldierx @AkalUstat @gurjit03 
Please let me know if there's anything else

gurjit03:
@yudhv veerji code looks good now from my side

yudhv:
Haven't added the tooltip option @saintsoldierx as it appears there can be several complications as mentioned [here](https://www.newline.co/@dmitryrogozhny/4-ways-to-show-tooltips-in-react-with-react-tooltip,-material-ui,-bootstrap,-or-semantic-ui--e3473190) if we implement it ourselves: 
`You need to consider positioning, responsive layouts, and various browsers support`
In-order to use `react-tooltip` I'd have to change the package.json which I feel should be left to more experienced individuals. For the interim, I've added pages in brackets in one of the earlier descriptions so a reader could know.

yudhv:
@saintsoldierx @gurjit03 Can we merge this?

saintsoldierx:
I am still looking into this @yudhv - i just need to verify some of the information and verbiage. Because if you look at it, the first raag in SGGSJ actually is Raag Asa (ang 8), although "chapters" of raags begin with Sri Raag, so want to make sure we provide correct info with the correct verbiage. Could you tell me the source of the information/text you have put?

yudhv:
@saintsoldierx (SikhWiki)[https://www.sikhiwiki.org/index.php/Index_of_Guru_Granth_Sahib] and some other sources opiniate that SGGS starts with the Sri Raga. The bani up til ang 13 is assumed to be Pre-Raga bani that encapsulates a more summarized version of the Sikh philosophy. That is why the Raags found from ang 8-13 are repeated again (ASA from 347, Gujari from 489, Dhanasari from 660). This repetition of Raags happens for no other Raags, but the ones between 8-13, which could signify that the bani therein has some kind of importance to it. The fact that So Dar and Sohila also happen to be part of Nitnem speaks to a similar view. That's why I went with the SikhWiki approach and not the one mentioned in Wikipedia (here)[https://en.m.wikipedia.org/wiki/Ragas_in_the_Guru_Granth_Sahib].

saintsoldierx:
@yudhv thanks for that - I am going to check with a few sources to make sure but I do agree with you too. I know wikipedia posts for the raags has been hijacked by a certain academy and they are pushing their own agendas.

tarunsingh5:
@saintsoldierx its commonly accepted that sri raag is the first official raag, the bani before that is Nitnem (jap ji sahib, rehraas sahib, sohila)

PR #1171, opened 2020-07-14
Adds shortcuts bar on left

Gauravjeetsingh:
Added a shortcut bar with a card design. 

**Currently, in this PR, we have the following colors:**
Similar to how we show the shabad lines.
image

**But I also liked giving light grey to inactive shortcuts**
image

### DIFF
--- a/src/js/components/BaaniLine.js
+++ b/src/js/components/BaaniLine.js
@@ -13,8 +13,8 @@ export default class BaaniLine extends React.PureComponent {
     larivaar: PropTypes.bool.isRequired,
     unicode: PropTypes.bool.isRequired,
     fontSize: PropTypes.number.isRequired,
-    lineHeight: PropTypes.number.isRequired,
     fontFamily: PropTypes.string.isRequired,
+    lineHeight: PropTypes.number,
     visraam: PropTypes.object,
   };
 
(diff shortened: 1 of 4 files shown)


saintsoldierx:
- If i tap off the slid out menu, it should close it.
- Clicking on Mool Mantr (for example) doesn't actually work. It doesnt show.
- This either needs to be aligned or make bg of slide out panel all white so it matches. 
image

- White on light blue does not work - although your screenshots show it differently? (Chrome/OSX)
image

Also on banis...
- Make sure every clickable link area/button is 44px in height at least. 
image

@Gauravjeetsingh

saintsoldierx:
After entering in PIN, bani controller not working on the heroku deployment.

Gauravjeetsingh:
@saintsoldierx it's working for me now. Sometimes there's a problem with the sync code. Can you please check again?

Gauravjeetsingh:
- [ ] If i tap off the slid out menu, it should close it.

Clicking on the toggle button "All Banis" or "Shortcuts" does close the sidebar if it's opened. Does this mean something else?

- [ ] Clicking on Mool Mantr (for example) doesn't actually work. It doesnt show.

We never showed the shortcuts on web side. We just showed it in the sttm-desktop. Do we want on the web as well?

- [ ] White on light blue does not work - although your screenshots show it differently? (Chrome/OSX)

Yeah, unable to reproduce on android and Linux. Will check mac.

Fixed the following:

- [x] This either needs to be aligned or make bg of slide out panel all white so it matches.

- [x] Make sure every clickable link area/button is 44px in height at least.

@saintsoldierx

PR #1179, opened 2020-07-16
Fix highlight scrolling

gurjit03:

Scroll was going to the next baani-line from the highlighted one. This PR fixes that bug


plug-n-play:
Observed this to be highlighting the searched line - https://khalis-herok-fix-highli-bi6pyh.herokuapp.com/shabad?id=2610&q=bhbgg&type=0&source=G&highlight=29479

Issue #1164, opened 2020-07-11
Make UX better to close Advanced Settings


plug-n-play:
As of now, when `Advanced` are open, it can be confusing to new users on how to close the Advanced settings
![image](https://user-images.githubusercontent.com/6245927/87217939-2b62c900-c36c-11ea-8c9d-43ad2030f1a7.png)

Recommendation:
1. Add a cross icon to close the Advanced Settings panel in addition to the existing option to close it
2. Change the UI state of `Advanced` control to distinguish between open and close states

gurjit03:
@saintsoldierx veerji, can you suggest UI state changes for `advanced` button ?

saintsoldierx:
For now, a X in top right corner @gurjit03

PR #1185, opened 2020-07-20
Update advance settings

gurjit03:

This PR is in reference to the issue #1164 


saintsoldierx:
@gurjit03 this is fine - but make it at least 34px width and height. i would prefer 44px but 34 will do for now.

PR #1207, opened 2020-08-17
Feature sehaj path mode

gurjit03:


AkalUstat:
its a little laggy...is there a way to maybe used a virtualized list? Also, what about if we end at like 5 angs later, how would one know what ang they ended on?

gurjit03:
@AkalUstat yeah haven't tried looking into the virtualized list, but yeah could give it a shot.
Ang are being updated in the URL, but yeah would be better if we could show it on the page itself somehow.

AkalUstat:
https://virtuoso.dev/ I like this for lists of unknown sized elements

PR #1217, opened 2020-08-26
Fix phonetic search

gurjit03:

- Fixes Phonetic search to allow W, I, O, U letters to entered in input
- Fixes automatic search type fallbacks to `Guru Granth Sahib Ji` bug.


src/js/util/gurbani/get-highlight-indices.ts
```diff
@@ -97,7 +97,7 @@ export const getHighlightIndices = (
     }
 
     case SEARCH_TYPES.ROMANIZED_FIRST_LETTERS_ANYWHERE: {
-      baaniWords = baaniWords.map(w => (w.startsWith('i') ? w.slice(1) : w));
+      // baaniWords = baaniWords.map(w => (w.startsWith('i') ? w.slice(1) : w));
```
plug-n-play, line 100: Do we want to keep this?

gurjit03: Just have made it as comment for reference

PR #1218, opened 2020-08-27
Feature index page tests

gurjit03:

image

### DIFF
--- a/cypress/integration/navigation/index.spec.js
+++ b/cypress/integration/navigation/index.spec.js
@@ -1,51 +1,7 @@
 import { raagIndices } from '../../../src/js/pages/GranthIndex/constants/raag-indices';
 
-const getRandomNumber = (max) => Math.floor(Math.random() * Math.floor(max));
-
-Cypress.Commands.add('checkGranthIndices', function checkGranthIndices({ granthIndex, indices, source }) {
-  const randomIndex = getRandomNumber(indices.length);
-
-  cy.get('.granthIndex')
-    .eq(granthIndex)
-    .find('tbody tr')
-    .as('granthRows')
-
-  cy.get('@granthRows')
-    .find('td a')
-    .then(anchors => {
-      anchors.each((idx, anchor) => {
-        // we just gonnna check 3 indexes
-        if (idx === 0 ||
-          idx === indices.length - 1 ||
-          idx === randomIndex) {
-          const ang = indices[idx];
-          const href = anchor.getAttribute('href');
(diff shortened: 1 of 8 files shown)


src/js/pages/GranthIndex/components/AmritKeertanIndex.tsx
```diff
@@ -11,7 +11,7 @@ interface IAmritKeertanProps {
 export const AmritKeertanIndex: React.FC<IAmritKeertanProps> = ({ data }) => {
 
   return (
-    <div>
+    <div className="granthIndex">
```
tsingh777, line 14: if this class just for the selector? if yes then prefer if `data-cy` like in HeaderMenu.

src/js/pages/GranthIndex/components/GranthIndex.tsx
```diff
@@ -75,7 +74,7 @@ export const GranthIndex: React.FC<IGranthIndexProps> = ({ location }) => {
               <main>
                 {Object.entries(raagIndices).map(
                   ([key, { name: granthName, source, indices }]) => (
-                    <React.Fragment key={key}>
+                    <div className="granthIndex" key={key}>
```
tsingh777, line 77: if this class just for the selector? if yes then prefer if data-cy like in HeaderMenu.

cypress/integration/navigation/index.spec.js
```diff
@@ -0,0 +1,134 @@
+import { raagIndices } from '../../../src/js/pages/GranthIndex/constants/raag-indices';
+
+const getRandomNumber = (max) => Math.floor(Math.random() * Math.floor(max));
+
+Cypress.Commands.add('checkGranthIndices', function checkGranthIndices({ granthIndex, indices, source }) {
```
tsingh777, line 5: should this go in `cypress/support/commands.js` ?

tsingh777: I've seen it where the index file under support contains common commands and then a file for each specific suite has custom commands for only that suite. It would make reading the test file easier

gurjit03: @tsingh777 thanks a lot veerji for the suggestion. Yeah since this was custom command so I just kept it enclosed but 100% agree with you that way it makes code easier to read.

tsingh777:
Before:
image
After:
image

plug-n-play (approved):

saintsoldierx:
@gurjit03 conflict

gurjit03:
@saintsoldierx resolved !

PR #1255, opened 2020-09-19
UI changes, Shahmukhi to right align, improvements to toast notifications

saintsoldierx:


package.json
```diff
@@ -114,6 +114,7 @@
     "express": "^4.17.1",
     "lodash.throttle": "^4.1.1",
     "marinate": "^0.1.5",
+    "postcss-root-to-json": "0.0.0",
```
tsingh777, line 117: where is this used?

saintsoldierx: Not sure, it was something added in automatically. https://developer.aliyun.com/mirror/npm/package/postcss-root-to-json

src/scss/_shabad-page.scss
```diff
@@ -73,6 +77,10 @@ blockquote {
   flex-grow: 1;
 }
 
+.mixed-view-baani-transliteration-shahmukhi {
+  float: right;
+}
+
```
gurjit03, line 83: @saintsoldierx can we try using flex-box properties here. Just for consistency and also float needs to be cleared as well,  otherwise generally they end up causing layout issues

gurjit03 (changes requested):

src/scss/_shabad-page.scss
saintsoldierx: image
 Would this be sufficient @gurjit03 ?

PR #1262, opened 2020-09-27
Fix mahankosh tooltip

gurjit03:

- Improves the Mahan Kosh tooltip implementation 
- Fixes multiple words being highlighted bug


saintsoldierx (approved):

saintsoldierx:
Generally is much better. At the moment, if a word doesn't have an entry, it goes bold for a split second. If we can improve that so it doesn't go bold at all, that'd be great. @gurjit03

gurjit03:
@saintsoldierx sure would aim for that indeed.

Issue #1268, opened 2020-10-03
Add a dropdown to select Bani length for SG


saintsoldierx:
Add ability to select Bani length for Sunder Gutka - default should be SHORT.

![image](https://user-images.githubusercontent.com/27791330/94983854-5ab44c80-04fb-11eb-9657-455fe02e18e5.png)

plug-n-play:
Working as expected.

@saintsoldierx Suggested change: IMHO it's better to have this option at Baani level rather than at Sundar Gutka level. For example, if user is reciting Choupai Sahib but realises he wants to read longer/shorter version while reciting it, he should be able to do it from read mode itself rather than switching back and forth and checking different versions - Short-to-Extra-Long. If we agree, we can create a new ticket for this

PR #1282, opened 2020-10-20
#1268 Feature sg baanis

gurjit03:


src/js/pages/SundarGutka/SundarGutkaHeader.tsx
```diff
@@ -0,0 +1,37 @@
...
+  sgBaaniLength,
+  setSgBaaniLength,
+}) => {
+  return (
+    <div className="sundarGutkaHeader">
+      <h2>{TEXTS.SUNDAR_GUTKA_HEADER}</h2>
+      <div className="sgBaanis">
+        <span className="sgBaanisInfoIcon" data-tip data-for="sgBaanisInfo">
+          <ReactTooltip backgroundColor={STTM_BLUE} place="top" id="sgBaanisInfo">
+          <span>Different version of Sundar Gutka Baanis</span>
```
saintsoldierx, line 19: Does this need to be in a language/parameter file?

gurjit03: @saintsoldierx we don't have any language file yet. But yeah eventually we would like to keep all the texts in there

PR #1286, opened 2020-10-28
Added script to run and record and upload cypress tests

plug-n-play:

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "start": "npm run server",
     "dev": "npm run watch",
     "cypress:open": "cypress open",
-    "cypress:record": "cypress run --record --key ad353245-3b1d-4adf-8dba-b7d7481263d1",
+    "cypress:record": "cypress run --record --key --env=cypress-sttm-web",
     "watch": "run-p watch:*",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -9,6 +9,7 @@
     "start": "npm run server",
     "dev": "npm run watch",
     "cypress:open": "cypress open",
+    "cypress:record": "cypress run --record --key ad353245-3b1d-4adf-8dba-b7d7481263d1",
```
tsingh777, line 12: security breach. should pull key from env variable

package.json
```diff
@@ -9,6 +9,7 @@
     "start": "npm run server",
     "dev": "npm run watch",
     "cypress:open": "cypress open",
+    "cypress:record": "cypress run --record --key --env=cypress-sttm-web",
```
tsingh777, line 12: check out https://docs.cypress.io/guides/guides/command-line.html#cypress-run-record-key-lt-record-key-gt 

think you just want this to be `cypress run --record` the env variable would be called `CYPRESS_RECORD_KEY`

tsingh777 (approved):

PR #1287, opened 2020-10-29
Ignore Cypress Screenshots and Videos

plug-n-play:

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -7,5 +7,6 @@
     "colors": {
       "sttmBlue": "rgb(1, 102, 155)"
     }
-  }
+  },
+  "projectId": "7y3f9z"
 }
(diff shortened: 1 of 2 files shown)


cypress.json
```diff
@@ -7,5 +7,6 @@
     "colors": {
       "sttmBlue": "rgb(1, 102, 155)"
     }
-  }
-}
\ No newline at end of file
+  },
+  "projectId": "7y3f9z"
```
tsingh777, line 11: What does this add? does this need to a secret?

saintsoldierx: @plug-n-play

saintsoldierx: I don't think this needs to be a secret 

https://docs.cypress.io/guides/dashboard/projects.html#Project-ID

Issue #1261, opened 2020-09-27
Add Hukamnama Audio 


saintsoldierx:
https://github.com/KhalisFoundation/hukamnama-json/pull/25

Add a way to play Hukamnama from Sri Harimandir Sahib audio from the Hukamnama page.

plug-n-play:
Don't see a control on UI to play Hukamnama audio

saintsoldierx:
This works - just missing data in DEV/STG

PR #1301, opened 2020-11-08
Feature hukamnama audio - #1261

gurjit03:

image


tarunsingh5:
i feel like the UI here is very confusing the left/right arrows change the day but the play button in between plays it, can we make that more clear somehow?

tarunsingh5:
are we storing audio for past days?

saintsoldierx:
No. Not at this time. The player will be removed for past days.

saintsoldierx:
image
I will be changing the layout so audio player moves to bottom. Rewind and forward buttons are pretty standard for this function.

PR #1326, opened 2020-12-13
Update hukamnama UI

gurjit03:


saintsoldierx:
Can we make it auto appear at bottom without them having to click on it IF they open any hukamnama? @gurjit03

gurjit03:
@saintsoldierx sure veerji

saintsoldierx:
So if i click on the Listen Today button it toggles the player which is fine. But then button stops working if pressed multiple times to toggle. Also, the player should have a title which says "Listen to Today's Hukamnama" and ideally a close button.

PR #1361, opened 2021-03-12
Add author filter

navdeepsingh:


common/api-urls-constants.js
```diff
@@ -12,5 +12,6 @@ module.exports = {
     LOCAL: '//stgapi.sikhitothemax.org/',
   },
   CEREMONIES: '//api.sikhitothemax.org/ceremonies/',
-  DOODLE: '//api.sikhitothemax.org/doodle/'
+  DOODLE: '//api.sikhitothemax.org/doodle/',
+  WRITERS: '//api.banidb.com/v2/writers'
```
bogas04, line 16: Add trailing slash to stay consistent

src/js/pages/Home/index.js
```diff
@@ -211,6 +215,64 @@ export default class Home extends React.PureComponent {
...
+                                    {writer.writerEnglish}
+                                  </option>
+                                ))
+                              ) : (
+                                writers?.map(writer => (
+                                  <option key={writer.writerID} value={writer.writerID}>
+                                    {writer.writerEnglish}
+                                  </option>
+                                ))
+                              )
```
bogas04, line 272: I think all this can be instead expressed as:

```js
writers?
  .filter(e => source
    ? SOURCE_WRITER_FILTER[source].includes(e.writerID)
    : true
  )
  .map(writer => (
    <option key={writer.writerID} value={writer.writerID}>
      {writer.writerEnglish}
    </option>
   ))
```

bogas04 (changes requested):
Left some comments. LGTM

src/js/pages/Home/index.js
navdeepsingh: Revised, as per need.

PR #1398, opened 2021-04-19
Feature/support ids query string

navdeepsingh:
Code for review.


server/utils/shabads/is-list-of-shabads.js
```diff
@@ -0,0 +1,3 @@
+export const isListOfShabads = object => {
```
gsin11, line 1: Somehow, I can't see the need of having this here, since you are checking property that is hardcoded here which may stop others to re-use it.

**Suggestion:**
I case you really need a function then either pass property name as param or keep this function inside create-metadata-from-response.js file only.

Happy to discuss further.

src/js/components/Autocomplete.js
```diff
@@ -112,7 +115,7 @@ class Autocomplete extends Component {
             .then(suggestions => {
 
               // if any suggestion exists, only then add this as final result item
-              if (isShowFullResults && suggestions.length > 0) {
+              if (isShowFullResults && suggestions.length) {
```
gsin11, line 118: Since you removed `array.length > 0` check from here, it will consider it true even when your array is empty?

src/js/components/Autocomplete.js
```diff
@@ -146,6 +149,12 @@ class Autocomplete extends Component {
     }
   }
 
+  handleAddShabads(e) {
+    e.preventDefault();
```
gsin11, line 153: If the type of this button is not `submit` then we don't need `e.preventDefault()`. Perhaps, it shouldn't submit button inside a form.

src/js/components/Autocomplete.js
```diff
@@ -157,6 +166,7 @@ class Autocomplete extends Component {
         isShowFullResults = false,
         value,
         searchOptions,
+        isHome,
```
gsin11, line 169: Is it me only who can see `isHome` is also part of state object or its just the rearranging of object keys?

package.json
```diff
@@ -133,13 +134,15 @@
...
+    "react-sortablejs": "^6.0.0",
     "react-toastify": "^6.0.9",
     "react-tooltip": "^4.2.9",
     "redux": "^4.0.5",
     "redux-debounced": "^0.5.0",
     "redux-mock-store": "^1.5.4",
     "redux-thunk": "^2.3.0",
-    "reselect": "^4.0.0"
+    "reselect": "^4.0.0",
+    "sortablejs": "^1.13.0"
```
gsin11, line 145: Does `react-sortablejs` needs `sortablejs` library along, thought it is already part of `react-sortablejs` itself?

src/js/components/MultipleShabadsDisplay/MultipleShabadsDisplay.tsx
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  // Update Local State {sortableState} after shabads get Updated
+  useEffect(() => {
+    setSortableState(shabads)
+  }, [shabads])
+
+  return (
+    <div className="multiple-shabads-display">
+      <h3>Multiple Shabads Display</h3>
```
gsin11, line 49: Not sure, but can we use the constant files to drive these string entries. I can see one here: `src/js/constants/texts.ts`?

src/js/components/ShabadContent/ListOfShabads.js
```diff
@@ -0,0 +1,171 @@
...
+ */
+class Shabad extends React.PureComponent {
+  /**
+   * @typedef {object} ShabadState
+   * @property {number} progress of vertical scroll
+   *
+   * @memberof Shabad
+   */
+  state = {
+    progress: 0,
```
gsin11, line 28: Where exactly you are using `progress` here?

src/js/components/ShabadContent/ListOfShabads.js
```diff
@@ -0,0 +1,171 @@
...
+                          isParagraphMode={false}
+                        />
+                      }
+                    </div>
+                  </div>
+                </React.Fragment>
+              )
+            })
+          }
+          {<ProgressBar />}
```
gsin11, line 163: When exactly this component will unmount?

src/js/pages/Home/index.js
```diff
@@ -58,6 +63,7 @@ export default class Home extends React.PureComponent {
    * Functional component
    */
   render() {
+    const { multipleShabads, setMultipleShabads, clearMultipleShabads, removeMultipleShabads, isHome } = this.props;
```
gsin11, line 66: I would suggest reconsidering this logic here, we are connecting the store to Homepage now which may cause re-renders of this page.
Please check if you can use these inside the `MultipleShabadsDisplay` component?

Can discuss it further.

gsin11 (commented):
I am done with my review.

src/js/components/Autocomplete.js
navdeepsingh: When suggestions array empty, which means array.length will return 0 which will false the if condition.

package.json
navdeepsingh: Yeah both need to install separately https://github.com/SortableJS/react-sortablejs

src/js/components/Autocomplete.js
navdeepsingh: Because event bubbles if dont use `e.preventDefault();` thats why.

src/js/pages/Home/index.js
navdeepsingh: Connected Individual Component instead giving whole control to Home Component.
Thanks!!

gsin11 (approved):
looks fine to me.

PR #1402, opened 2021-04-24
Feature/1400 accessibility enhancements 2

gsin11:
- add missing lang attribute on `<html>` tag
- add accessible color for links on non-dark mode
- add aria-label for empty text CTAs
- add text to blank link

## No more Errors and contrast issues on help page
help_page

### Pending (can be part of new PR)
- We should add language attribute wherever we are using non-English language, see example below:
`<span lang="pa" xml:lang="pa">Punjabi text</span>`

### DIFF
--- a/server/template.js
+++ b/server/template.js
@@ -19,9 +19,9 @@ try {
 <html lang="en">
 TODO: language attribute is hard coded for now, make it dynamic when actual language support is added.
 */
-export default ({ url, bodyClass, title, description }) => marinate`
+export default ({ url, bodyClass, title, description, language }) => marinate`
 <!DOCTYPE html>
-<html lang="en">
+<html lang="${language}">
 <head>
   <title>${title}</title>
   <link rel="shortcut icon" type="image/x-icon" href="/favicon.ico" />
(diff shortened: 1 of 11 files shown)


server/template.js
```diff
@@ -17,7 +17,7 @@ try {
 
 export default ({ url, bodyClass, title, description }) => marinate`
 <!DOCTYPE html>
-<html>
+<html lang="en">
```
tsingh777, line 20: we do want to support localization at some point. would we change this programmatically or use a variable we read from somewhere?

gsin11: Thanks for suggesting it @tsingh777, We definitely need to have such a thing in place for longer support, but before adding a variable we need to decide on a URL pattern for language support. Not sure if we have something already in place?
**Example**
`sikhitothemax.org/en/`
`sikhitothemax.org/en-US/`

However, I added a TODO comment for the future.

tsingh777: Not sure if we pattern yet for localization. Some options:
- Add a selector somewhere on the UI and store in a cookie/localstorage and read from that.  <- my preference 
- auto detect from browser locale 

Might need to refactor a little bit so we can use something like https://www.npmjs.com/package/react-intl or https://github.com/lingui/js-lingui

gsin11: thanks for your suggestion @tsingh777 !
I added a cookie to save the language, and in case it is not available fallback will go to "en" which is also defined as constant.

PR #1409, opened 2021-05-02
feat(accessibility): improved accessibility in right rail

gsin11:
- Right is being mounted on request instead of keeping it always on-screen hidden state.
- Replaced `<span>` with `<button>` for better accessibility and keyboard controls.
- Added focus state (similar to hover).

### Anything else?
We have react-modal in place, please use it for better performance and accessibility.

### DIFF
--- a/src/js/components/ControlsSettings/ControlsSettings.tsx
+++ b/src/js/components/ControlsSettings/ControlsSettings.tsx
@@ -48,6 +48,7 @@ const ControlsSettings = (props: any) => {
       VISRAAM.SOURCE_CLASS(visraamSource),
       VISRAAM.TYPE_CLASS(visraamStyle)
     );
+    wrapperRef.current.focus();
   }, [visraams, visraamSource, visraamStyle])
 
   const renderIcon = (itemName: any) => {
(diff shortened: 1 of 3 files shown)


navdeepsingh (approved):

src/js/components/ControlsSettings/ControlsSettings.tsx
```diff
@@ -140,21 +141,22 @@ const ControlsSettings = (props: any) => {
...
-              <span className="settings-text">{settingsObj.label}</span>
-              <div className="flex-spacer" />
-              <span className="settings-chevron-icon">
-                <SolidArrowRight />
-              </span>
-            </div>
-          )}>
+          <Collapsible
+            trigger={(
+              <button className="settings-item active-setting">
```
navdeepsingh, line 151: @gsin11 see if we can add `aria-expanded` too for collapsible items here?

gsin11: Yeah I was trying to add that but this plugin doesn’t have support for it, can we take this in next ticket?

gsin11: Or May be, I can remove this module and use my own?

PR #1652, opened 2023-06-02
Add feature "Ask a question" using semantic search API

Gauravjeetsingh:

Used the /shabads endpoint of banidb API and parsed the data to only what's required. We can update it once we have a separate endpoint to load just the verses we need.

Here is it in action:
gurbanibot1
gurbanibot2

Deployment link: https://khalis-herok-gurbanibot-6k71ha.herokuapp.com/


src/js/components/Autocomplete.js
```diff
@@ -242,7 +243,7 @@ class Autocomplete extends Component {
 
     return (
       <Fragment>
-        {suggestionsListComponent}
+        {!isChatBot && suggestionsListComponent}
```
mandeepryaz, line 246: In case the condition is false, it will return the empty Fragment? Is that expected?

src/js/pages/Search/index.js
```diff
@@ -20,8 +21,58 @@ export default class Search extends React.PureComponent {
...
+
+  setSearchUrl() {
+    const { q, type, source, offset, writer } = this.props;
+    const isChatBot = type === SEARCH_TYPES.ASK_A_QUESTION;
+    let semanticInfo;
+
+    if (isChatBot) {
+      const semanticApi = encodeURI(`${GURBANIBOT_URL}search/?query=${q}`);
+      const semanticReq = fetch(semanticApi).then((response) => response.json());
+      semanticReq.then((semanticData) => {
```
mandeepryaz, line 38: Should we be having some kind of error handling here?

src/js/pages/Search/index.js
```diff
@@ -20,8 +21,58 @@ export default class Search extends React.PureComponent {
...
+        const shabadIdList = semanticData.results.map((dataObj) => {
+          const { ShabadID, VerseID } = dataObj.Payload;
+          this.state.verseIdList.push(VerseID);
```
mandeepryaz, line 41: Can we push to some other array here and update the state just once after the map?

mandeepryaz (changes requested):

src/js/pages/Search/index.js
```diff
@@ -20,8 +21,58 @@ export default class Search extends React.PureComponent {
...
+          return ShabadID;
+        });
+
+        semanticInfo = {
+          pageResults: shabadIdList.length,
+          TotalResults: shabadIdList.length,
+          pages: {
+            page: 1,
+            resultsPerPage: 20,
```
mandeepryaz, line 50: Does it make sense to add this to some config file or you think it's better suited here?

src/js/pages/Search/utils/parseSemantic.js
```diff
@@ -0,0 +1,17 @@
+export const parseSemanticData = (semanticData, verseIdList) => {
+  const parsedData = {};
+  parsedData.verses = semanticData.shabads.map((shabadObj) => {
+    const { shabadInfo } = shabadObj;
+    shabadObj.verses = shabadObj.verses.reduce((filteredVerses, verseObj) => {
+      if (verseIdList.includes(verseObj.verseId)) {
+        verseObj.source = shabadInfo.source;
+        verseObj.raag = shabadInfo.raag;
+        verseObj.writer = shabadInfo.writer;
```
mandeepryaz, line 9: Probably this can be destructured.

mandeepryaz (changes requested):

src/js/components/Autocomplete.js
```diff
@@ -77,10 +78,10 @@ class Autocomplete extends Component {
...
   onMouseDown = e => {
     e.stopPropagation();
-    (this.state.showSuggestions && !this.wrapperRef.current.contains(e.target)) &&       
-    this.setState({
-      showSuggestions: false,
-    });    
+    (this.state.showSuggestions && !this.wrapperRef.current.contains(e.target)) &&
+      this.setState({
+        showSuggestions: false,
+      });
```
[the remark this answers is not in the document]
Gauravjeetsingh: From what I saw, this entire codebase is currently on class-based components. This would be a huge refactor if we plan to do.

src/js/pages/Search/index.js
Gauravjeetsingh: This will come from banidb API, we won't need to write this manually here, its just a temporary code, until we get that endpoint in production.

src/js/components/Autocomplete.js
Gauravjeetsingh: Thank you ji, this should be fixed with recent commit.

src/js/pages/Search/index.js
Gauravjeetsingh: Added a try.. catch on this, I believe that would be sufficient here.

src/js/pages/Search/index.js
Gauravjeetsingh: Done, thanks ji!

src/js/pages/Search/utils/parseSemantic.js
Gauravjeetsingh: Yes, fixed.

src/js/components/Autocomplete.js
Gauravjeetsingh: I reverted this back, react gives an error if we don't return that empty fragment.

Gauravjeetsingh:
Home page and disclaimer message
gurbanibot-welcome

Pagination in search results
gurbanibot-scroll

Gauravjeetsingh:
In order for this code to work, we need to merge the BaniDB endpoint and deploy it to production
https://github.com/KhalisFoundation/banidb-api/pull/268

saintsoldierx:
image
I see this issue of the dropdown not appearing correctly on subpages. @Gauravjeetsingh

saintsoldierx:
image

I would like the "let us know" to be linked to our contact page (https://support.khalisfoundation.org/support/tickets/new)

saintsoldierx:
NEW Click here to get your questions answered from the Gurbani.

I would like this to read:

**NEW** Get your questions answered by our AI bot! Try it now. 

Try it now should be the link, not underlined. AI is quite a buzzword right now so I think it will attract people.

Gauravjeetsingh:
@saintsoldierx veerji, couldn't reproduce the dropdown issue, is it happening everytime? Or at a certain screen size?

Done with other two changes.

PR #1705, opened 2024-06-25
Moving semantic search API to Google cloud run

Gauravjeetsingh:
With google cloud run, the API is hosted at gurbanichatbot.sikhitothemax.org


gurjit03 (approved):
LGTM !

Issue #1111, opened 2020-06-10
ignore - in sync and controller code


ManjotS:
**Describe the bug**
when I type code - is autofilled. If I type it I end up with RRF--EEH

Have the interface ignore the key '-'

gurjit03:
@saintsoldierx @Gauravjeetsingh is this bug still exists or we can close the ticket otherwise.

Gauravjeetsingh:
@gurjit03 veerji, this issue still exists in production.

gurjit03:
@Gauravjeetsingh okay I see, already in our pipeline to get it fixed or it's something very minor it doesn't really bother sangar for now?

Gauravjeetsingh:
I think that would be @saintsoldierx veerji's call.

gurjit03:
@saintsoldierx ?

saintsoldierx:
@Gauravjeetsingh this still needs fixing and I can imagine it's annoying for people using regularly or new users too so let's look at getting this done soon.

Gauravjeetsingh:
@gurjit03 veerji, you got this or let me know if you want me to look over it.

Gauravjeetsingh:
Adding context behind this issue here:

The bani controller code is in format ABC-XYZ

And when we enter it on its connect page, https://www.sikhitothemax.org/control, if we type the first three letters of the sync code, it automatically adds the hyphen (-) 

But if we add hyphen ourselves, then it shows two hyphens. We need to disable manual hyphen from that input. Or, if the manual hyphen is added, don't add another hyphen after it.

PR #1725, opened 2024-07-24
fix hyphen issue for sync and controller code

yadvirkaur:

#1111

### DIFF
--- a/src/js/pages/WebController/index.js
+++ b/src/js/pages/WebController/index.js
@@ -14,7 +14,6 @@ import { Stub } from '../Search/Layout';
 import ControllerShabad from '@/pages/WebController/shabad';
 import { versesToGurbani } from '@/util';
 import ShabadControls from '@/components/ShabadControls';
-import { isMobile } from 'react-device-detect';
 export default class WebControllerPage extends React.PureComponent {
   constructor(props) {
     super(props);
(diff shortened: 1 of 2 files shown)


src/js/pages/WebController/index.js
```diff
@@ -274,26 +273,18 @@ export default class WebControllerPage extends React.PureComponent {
                   pattern="[A-Z,a-z]{3}-[A-Z,a-z]{3}"
                   onKeyUp={e => {
                     const typedValue = e.currentTarget.value;
-                    if (isMobile) {
```
gurjit03, line 277: @yadvirkaur I can see there is difference in the logic in Mobile and Desktop. Is it something reconsidered now?

src/js/pages/WebController/index.js
```diff
@@ -274,26 +273,18 @@ export default class WebControllerPage extends React.PureComponent {
...
+                    const typedChar = e.key;
+                    const hasHyphen = typedValue.includes('-');
+                    const parsedValue = typedValue.match('^[A-Z,a-z]{3}');
+                    const isParsedValueExist = parsedValue
+                      ? parsedValue[0] === typedValue
+                      : false;
+                    if (isParsedValueExist && typedChar !== 'Backspace') {
+                      e.currentTarget.value = typedValue + '-';
+                    } else if (typedChar === '-' && hasHyphen) {
+                      e.currentTarget.value = typedValue.replace(/-+/g, '-');
```
gurjit03, line 285: @yadvirkaur I might be wrong here but we are replacing `/-+/g` both `-` and `+` aint? Or does it represents `-` n number of times? if so, can you please update implementation a bit to be more clear. It can cause confusion.

gurjit03 (commented):
Great work @yadvirkaur ji 🔥. Can you please resolve the below comments and also please add a unit test for this function and/or component. 
And one more request please whenever you work/touch a file lets convert it to Typescript immediately to get better type support and enhance DX.

src/js/pages/WebController/index.js
yadvirkaur: **Previously:**

**For the Bani Controller:**

- **Mobile Devices:** The code checked if the user had typed four characters without a hyphen and then inserted a hyphen after the third character.

- **Other Devices:** The code added a hyphen automatically after the third character.

**For Sangat Sync:**

The code for all devices added a hyphen automatically after the third character.

**Now:**
I've now unified the logic for the Bani Controller to ensure consistent behaviour across all devices. The code now adds a hyphen automatically after the third character for all devices and ensures that no extra hyphens are added.

gurjit03: @yadvirkaur cool, thanks for the explainations, I totally got it. It's just I am unaware of these changes, if these are the changes requested lets do it, and also I would say lets please make sure these are the part mentioned in the ticket, easier to review and followup. 
Otherwise I would request to kindly confirm, as I am sure there might be good reason for this seperation.

src/js/pages/WebController/index.js
yadvirkaur: The expression `(/-+/g, '-')` here matches one or more hyphens and replaces them with a single hyphen. It's not related to the `+` character itself. To make it clearer, I can update it to` (/[-]+/g, '-')`. Let me know if that works!

src/js/pages/WebController/index.js
yadvirkaur: Totally! 
The mobile-specific logic for the Bani controller is fine as it is since it adds a hyphen after 4 characters only if the user hasn't already added one, so no risk of double hyphens there.

However, this can cause confusion between the Bani Controller and Sangat Sync since they handle hyphens differently for mobile devices.

@saintsoldierx @Gauravjeetsingh veerji please confirm if we need separation of logic or if we can have a unified logic.

src/js/pages/WebController/index.js
gurjit03: @yadvirkaur yes that works perfectly fine

src/js/pages/WebController/index.js
Gauravjeetsingh: I don't remember the reason for separate logic, but if its working with the unified logic, let's keep that. Realistically it makes sense to have same logic for both mobile and desktop browsers.

gurjit03 (approved):

PR #1769, opened 2024-11-04
updates hukamnama url

gurjit03:
In this PR, we have updated the source of the hukamnama audio to be from same source as sgpc.


JapnitS (approved):
LGTM

PR #1792, opened 2025-05-13
Remove node_env from Dockerfile

Gauravjeetsingh:
Removing env variable `NODE_ENV` from `Dockerfile` as the same dockerfile would be used for dev deployments.


Gauravjeetsingh:
no longer needed

PR #1798, opened 2025-06-02
Fix the autoplay not working

Gauravjeetsingh:
**The Issue:**
The auto-scroll wasn't working properly on Windows Chrome because of how we were tracking scroll positions
The original code was incrementing _nextScrollPosition without considering the actual current scroll position:
```javascript
this._nextScrollPosition += movement;  // Problematic: Lost track of actual position
```

**The Fix:**
Changed how we calculate the next scroll position to always use the actual current position:
```javascript
this._nextScrollPosition = scrollY + movement;  // Fixed: Always based on current position
```

Also, removed the firefox specific code, as that was causing autoplay to break.

Sidenote:
Restructured the code a bit for code readablity and fixed linter errors as well.

### DIFF
--- a/src/js/components/AutoScrollControl/AutoScrollControl.tsx
+++ b/src/js/components/AutoScrollControl/AutoScrollControl.tsx
@@ -172,9 +172,11 @@ class AutoScrollControl extends React.PureComponent<
         2
       );
 
-      // Update next position and scroll
-      this._nextScrollPosition = scrollY + movement;
-      window.scrollTo(0, Math.round(this._nextScrollPosition));
+      if (scrollY >= Math.round(this._nextScrollPosition)) {
+        // Update next position and scroll
+        this._nextScrollPosition = scrollY + movement;
+        window.scrollTo(0, Math.round(this._nextScrollPosition));
+      }
 
       this._interval = requestAnimationFrame(this.handleAutoScroll);
     }
--- a/src/js/components/AutoScrollControl/AutoScrollControl.tsx
+++ b/src/js/components/AutoScrollControl/AutoScrollControl.tsx
@@ -3,14 +3,14 @@ import { connect } from 'react-redux';
 import { Range } from 'react-range';
 
 import { toFixedFloat } from '../../util/numbers';
-import { Pause, Play,Loop } from "../../components/Icons/controls";
+import { Pause, Play, Loop } from '../../components/Icons/controls';
 import { setAutoScrolling } from '@/features/actions';
 
 interface IReduxStateAsProps {
   isAutoScrolling?: boolean;
 }
 interface IReduxDispatchProps {
-  setAutoScrolling?: (action: boolean) => {}
+  setAutoScrolling?: (action: boolean) => {};
 }
 
 interface IAutoScrollControlState {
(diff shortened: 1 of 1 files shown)


src/js/components/AutoScrollControl/AutoScrollControl.tsx
```diff
@@ -247,34 +275,53 @@ class AutoScrollControl extends React.PureComponent<IAutoScrollControlProps, IAu
...
+});
 
 const mapDispatchToProps: IReduxDispatchProps = {
-  setAutoScrolling
-}
+  setAutoScrolling,
+};
 
 // TODO: finding fix for proper typing this
-export default connect<{}, {}, IAutoScrollControlProps, IAutoScrollControlState>(mapStateToProps as any, mapDispatchToProps)(AutoScrollControl as any);
```
gurjit-distribusion, line 280: Also would be great if we can add another ticket to convert it to functional code.

src/js/components/AutoScrollControl/AutoScrollControl.tsx
```diff
@@ -97,74 +110,75 @@ class AutoScrollControl extends React.PureComponent<IAutoScrollControlProps, IAu
...
+      );
 
-      // We are having minimum scroll per pixel + adding the extra dynamic pixel movement based on slider value.
-      const movement = toFixedFloat((0.1 + ((scrollingSpeed / 100) * AutoScrollControl.thresHold)), 2);
-      // Only allow the scrolling if we have surpassed previous scrolls or if it's firefox browser.
-      if (this._isFirefoxAgent || scrollY >= Math.round(this._nextScrollPosition)) {
-        this._nextScrollPosition += movement;
-        window.scrollTo({ left: 0, top: Math.round(this._nextScrollPosition)});
-      }
+      // Update next position and scroll
```
gurjit03, line 175: @Gauravjeetsingh In the previous logic, we had an if condition to prevent scrollY updates when the new scroll position surpassed the previous one. I've reintroduced that check to avoid unnecessary scroll position updates.
It's a small detail, but it's good to keep track of these changes for consistency.

gurjit03 (approved):

src/js/components/AutoScrollControl/AutoScrollControl.tsx
```diff
@@ -247,34 +275,53 @@ class AutoScrollControl extends React.PureComponent<IAutoScrollControlProps, IAu
...
 
 const mapDispatchToProps: IReduxDispatchProps = {
-  setAutoScrolling
-}
+  setAutoScrolling,
+};
 
 // TODO: finding fix for proper typing this
-export default connect<{}, {}, IAutoScrollControlProps, IAutoScrollControlState>(mapStateToProps as any, mapDispatchToProps)(AutoScrollControl as any);
+export default connect<
```
gurjit03, line 319: Also, it's good if we can have a ticket to conver this to functional code.

gurjit03 (approved):

src/js/components/AutoScrollControl/AutoScrollControl.tsx
Gauravjeetsingh: Fixed, added that condition back

src/js/components/AutoScrollControl/AutoScrollControl.tsx
Gauravjeetsingh: https://github.com/KhalisFoundation/sttm-web/issues/1800

PR #1822, opened 2025-09-15
Add a loader when opening Ask a question popup

Gauravjeetsingh:
Adds a loader when opening a Ask a question popup

https://github.com/user-attachments/assets/2b52227b-1b88-4ccd-a719-63b682b39776


amitojsingh (approved):
Looks good to me

PR #1851, opened 2026-02-12
integrate playwright

mandeeps708:
- Add 2 tests on the homepage
- Update code to add uniquely ids to few elements
- Ignore e2e tests directory in Jest config
- Add GitHub action
- Update README


Gauravjeetsingh (commented):
@mandeeps708 veerji, the implementation looks good to me. 

I do have a concern on why are we integrating another e2e testing in this app. It already has few cypress tests. As far as I understand, both perform the same task, with the key difference in their implementation. 

If there is already a green light on **using playwright instead of cypress**, can we remove the cypress integration, and migrate those tests to playwright as well.

indersinghkhalis:
@Gauravjeetsingh The greenlight for playwright was given by me. To keep scope of volunteer contributions bite-sized, the further migration of cypress tests can be part of future PRs.

Gauravjeetsingh (approved):

Gauravjeetsingh:
Alright, the current PR is good to merge. I created a github issue for future PRs regarding playwright tests here:
https://github.com/KhalisFoundation/sttm-web/issues/1852

PR #1877, opened 2026-08-04
fix: false offline screen from unreliable navigator.onLine

divyanshugarg36:
Some users (seen in Chrome) get navigator.onLine === false while actually connected, which permanently pinned the app on the offline screen — the online event only fires on a transition, so a false initial value never self-corrected.

- Start optimistic (store online: true) instead of trusting navigator.onLine.
- On mount, only if navigator.onLine reports offline, confirm with a real same-origin request (verify-online util) and set online from the result.
- Verify the offline event the same way before showing the offline screen.

### DIFF
--- a/src/js/util/verify-online.ts
+++ b/src/js/util/verify-online.ts
@@ -1,15 +1,23 @@
+// Probe the app's data API health endpoint (banidb `/health`, which returns
+// `{ ok: true }` with permissive CORS) and trust the actual response, not mere
+// request completion. This confirms the backend the app depends on is really
+// reachable — a captive portal or a same-origin SPA fallback can't fake it.
+const apiBase =
+  (typeof API_URL === 'string' && API_URL) || 'https://api.banidb.com/v2/';
+const HEALTH_URL = `${apiBase}health`;
+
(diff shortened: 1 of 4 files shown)


src/js/util/verify-online.ts
```diff
@@ -0,0 +1,18 @@
+export async function verifyOnline(timeoutMs = 5000): Promise<boolean> {
+  if (typeof window === 'undefined' || typeof fetch === 'undefined') return true;
+
+  const controller = new AbortController();
+  const timer = setTimeout(() => controller.abort(), timeoutMs);
+  try {
+    await fetch(`/favicon.ico?_=${Date.now()}`, {
```
Gauravjeetsingh: This might give us false positive. How about using the health endpoint in the app, or use the one provided in API? 

And returning value based on an actual response, instead of just a request completion.

Gauravjeetsingh (changes requested):

src/js/util/verify-online.ts
divyanshugarg36: Yes, I updated it to ping banidb/health endpoint

Gauravjeetsingh (approved):

PR #1883, opened 2026-08-19
fix: keep Mahankosh tooltip open during autoscroll

### LINKED ISSUE #1771: Autoscroll breaks mahankosh tooltip modal
**Describe the bug**
The Mahankosh tooltip is immediately hidden when the autoscroll mode is enabled. This makes it impossible to interact with or read the tooltip.

**To Reproduce**
1. Navigate to [SikhiToTheMax Hukamnama](https://www.sikhitothemax.org/hukamnama?date=2024/11/25).
2. Click on Settings and enable Autoscroll Mode.
3. Also, enable the Mahankosh Tooltip option.
4. Click on any word to open the tooltip.

**Expected behavior**
The Mahankosh tooltip should remain visible when a word is clicked, even in autoscroll mode, until the user closes it or interacts with another word.

radiator13:
Fixes #1771

## Problem
Enabling Autoscroll Mode hid the Mahankosh tooltip as soon as a word was clicked.

## Cause
`MahankoshTooltip` set `globalCloseEvents.scroll: true` and `closeEvents.mouseleave: true`. Autoscroll both scrolls the page and moves the word out from under the cursor.

## Change
While `isAutoScrolling` is true, do not close on scroll or mouseleave. Click-outside and Escape still close the tooltip.

## Test
1. Open a hukamnama/shabad
2. Enable Autoscroll Mode and Mahankosh Tooltip
3. Click a word
4. The tooltip should stay visible until you dismiss it or click another word

### DIFF
--- a/src/js/components/Baani/Baani.js
+++ b/src/js/components/Baani/Baani.js
@@ -1095,6 +1095,7 @@ class Baani extends React.PureComponent {
         {selectedWord && gurbaniLineInfo && selectedWordIndex > -1 && (
         <MahankoshTooltip
           isTooltipOpen={isMahankoshTooltipActive}
+          isMahankoshTooltipActive={isMahankoshTooltipActive}
           clearMahankoshInformation={this.clearMahankoshInformation}
           tooltipId="mahankoshTooltipHighlightSearchResult"
           gurbaniWord={selectedWord}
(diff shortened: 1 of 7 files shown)


radiator13:
Follow-up commit `30f0ee7`: removed the redundant `isTooltipOpen` prop — Baani was passing the same redux value twice under two names. `isMahankoshTooltipActive` (required again) is now the single source for `isOpen`. No behaviour change; helper test suites still pass.

PR #1884, opened 2026-08-19
fix: copy Unicode when selecting a GurbaniLipi pankti

### LINKED ISSUE #1802: When copying a pankti, it should always copy the unicode text
If we click the copy to clipboard icon, it would copy unicode. But if we select text and press ctrl/cmd + c, it copies the non-unicode.

Image

This icon appears when we select a pankti

radiator13:
Fixes #1802

## Problem
The copy-icon copied Unicode. Selecting a pankti and pressing Ctrl/Cmd+C copied ASCII/GurbaniLipi.

## Cause
With Unicode display off, the DOM text is GurbaniLipi. Native copy uses that text.

## Change
- Store each line's unicode on `data-unicode-verse`
- On copy from a `.gurlipi` / `.gurlipi-reading-mode` line, write that unicode to the clipboard
- Leave Unicode-mode copies and translation copies unchanged

## Test
1. Open a shabad with Unicode display **off**
2. Select a pankti and press Ctrl/Cmd+C
3. Paste — it should be Gurmukhi Unicode, not ASCII

### DIFF
--- a/src/js/components/Baani/Baani.js
+++ b/src/js/components/Baani/Baani.js
@@ -33,6 +33,7 @@ import {
   steekMap,
   hindiTranslationMap,
   isShowSehajPaathModeRoute,
+  getUnicodeCopyText,
 } from '@/util';
 import { MahankoshContext } from '@/context';
 import { changeAng, prefetchAng } from './utils';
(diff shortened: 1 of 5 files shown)


radiator13:
Follow-up commit `2d80661` hardens the copy interception around edge cases:

- **Multi-pankti selections** now join every selected line's unicode (document order, de-duplicated) instead of copying only the anchor line.
- **Collapsed / empty selections** no longer hijack Ctrl+C — native behaviour preserved.
- **Mixed or non-GurbaniLipi selections** fall back to the native copy unless the selection starts *or* ends inside a `.gurlipi` / `.gurlipi-reading-mode` line (translations-only copies untouched).
- Util signature simplified to `getUnicodeCopyText({ unicodeMode, selection })`; unit tests extended to cover all of the above (verified passing against jsdom).

